PR #41, opened 2014-12-27
Black UI theme

mzyy94:
Here, a "black UI theme" pull-request. 
I'll make more improvements of black UI.


mzyy94:
Thank you for your review. At this moment, I don't have both machines, running GameStream on Windows and running Xcode on OS X. So, I'll fix the code in a week.

phantom-playR:
I made that for replace the "no image" picture. And it's possible to change the font ? (Here it's "Roboto")
new test

PR #50, opened 2015-01-05
Black UI theme v2

mzyy94:
This pull request has following improvements:
- Show an app title when the image of the app isn't available.
- Added/set Roboto font family.
- Change the style of the segmented control.
- Fixed some bugs.

(The Roboto font is licensed under Apache License v2.0. If there isn't license inconsistency, please merge this pull-request.)


dwaxemberg:
I read through the license and it seems as long as we provide a copy of that license (which you did), we can do with this as we please. I will merge this request later today.

PR #215, opened 2016-04-07
Drag and drop

Mimiste:
Possibility to initiate a drag and drop

- Touch for 2 second to initiate a drag
- Move the cursor as usual
- Tap to drop


Limelight/Input/StreamView.m
```diff
@@ -49,11 +52,25 @@ - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
         UITouch *touch = [[event allTouches] anyObject];
         touchLocation = [touch locationInView:self];
         touchMoved = false;
+        dragTimer = [NSTimer scheduledTimerWithTimeInterval:2.0
```
cgutman: Can we get another touchesBegan before touchesEnded? We don't want to leave a timer dangling.

Limelight/Input/StreamView.m
```diff
@@ -49,11 +52,25 @@ - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
...
+                                                     target:self
+                                                   selector:@selector(onDragStart:)
+                                                   userInfo:nil
+                                                    repeats:NO];
+    }
+}
+
+- (void)onDragStart:(NSTimer*)timer {
+    if (!touchMoved && !isDragging){
```
cgutman: It looks like if this code fires for a 2 finger drag, the left button will never be released.

Mimiste: I corrected this part to start the timer only if you touch with one finger.

But, do you have something for testing the app ? 
My problem is each time i want to test something, i have to reboot the computer under windows to launch geforce experience..

dwaxemberg:
I tried this out and it works pretty well, but 2 seconds is a very long time to not move your finger. Can we reduce that to maybe 1 second?

Mimiste:
You are right. After 1 week of using it's clearly too long. I changed it to 1 second

PR #216, opened 2016-04-12
Rewrote the Controller class in Swift

Aghassi:
- Rewrote the `Controller` class in Swift.
- Removed old `Controller` files that were Objective-C
- Created Bridge-Header file for future use
- Enabled `Defines Modules` in build settings to allow for Swift use in Objective-C 
- Created `MoonlightUnitTests` target for future testing.
- Wrote `ControllerUnitTests` for basic checks against the newly written `Controller` class.


Limelight/Input/ControllerSupport.m
```diff
@@ -8,9 +8,12 @@
 
 #import "ControllerSupport.h"
 #import "OnScreenControls.h"
-#import "Controller.h"
+//#import "Controller.h"
```
dwaxemberg: Can you remove this entirely instead of just commenting it out?

Limelight/Input/OnScreenControls.m
```diff
@@ -8,9 +8,12 @@
 
 #import "OnScreenControls.h"
 #import "ControllerSupport.h"
-#import "Controller.h"
+//#import "Controller.h"
```
dwaxemberg: Same here, remove the line.

Aghassi: Just pushed new commit, done.

PR #309, opened 2018-03-06
Added support for the remote streaming optimizations in moonlight-common

FelixKratz:
I thought it would be a good idea to add support for the remote streaming optimizations. With this implementation I fixed the packet size for remote streaming and added a toggle for this to the settings view.
Also, I deactivated the bitrate multiplier for HEVC streams in remote streaming mode, because during streaming remotely the bandwidth will most certainly be known. This will result in a higher streaming quality.

### DIFF
--- a/Limelight/Database/DataManager.h
+++ b/Limelight/Database/DataManager.h
@@ -14,7 +14,8 @@
 
 @interface DataManager : NSObject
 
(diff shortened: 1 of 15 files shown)


cgutman:
Merged as part of #311

PR #311, opened 2018-03-17
Port for macOS

FelixKratz:
This commit brings a native mac port that shares a lot of code with the ios version. There is still much to do but it already works. In the current state the mac version supports HEVC playback on macOS 10.13 and later. I also added support for DualShock 3 Controllers, because these pair natively with macOS and i had one laying around for testing.
Some things that are still missing:

- A nice GUI, along with the implementation of a better host and game selection
- GC & IOHID Controller mixing
- Some key mappings

But why could anyone use a native mac client?
- I observed a nice boost in picture quality at the same bitrate (mainly because of HEVC, i think).
- The performance of a native client written in C should bring down latency and increase the battery life during remote streaming.

This commit also contains #309


Limelight/Stream/Connection.m
```diff
@@ -310,63 +315,86 @@ -(void) terminate
...
-    
+
+
+#if TARGET_OS_IPHONE
     // RFI doesn't work properly with HEVC on iOS 11 with an iPhone SE (at least)
     _drCallbacks.capabilities = CAPABILITY_REFERENCE_FRAME_INVALIDATION_AVC;
-    
+#else
+    // It works on macOS (on my MBP atleast)
+    _drCallbacks.capabilities = CAPABILITY_REFERENCE_FRAME_INVALIDATION_HEVC;
```
cgutman: Did you test with packet loss scenarios?

You should also specify CAPABILITY_REFERENCE_FRAME_INVALIDATION_AVC here if relevant.

Limelight/Stream/Connection.m
```diff
@@ -310,63 +315,86 @@ -(void) terminate
...
         // FIXME: Disabled due to incompatibility with iPhone X causing video
         // to freeze. Additionally, RFI is not supported so packet loss recovery
         // is worse with HEVC than H.264.
+
+        //Streaming with a limited bandwith will result in better quality with HEVC
         //_streamConfig.supportsHevc = VTIsHardwareDecodeSupported(kCMVideoCodecType_HEVC);
     }
+#else
+    if (@available(macOS 10.13, *)) {
+        _streamConfig.supportsHevc = true;
```
cgutman: We should still use VTIsHardwareDecodeSupported() here, since not all Macs with 10.13 will have hardware HEVC decode, even if they have a software fallback.

Limelight/ViewControllers/SettingsViewController.m
```diff
@@ -66,6 +67,12 @@ - (void)viewDidLoad {
     [self.bitrateSlider setValue:(_bitrate / BITRATE_INTERVAL) animated:YES];
     [self.bitrateSlider addTarget:self action:@selector(bitrateSliderMoved) forControlEvents:UIControlEventValueChanged];
     [self updateBitrateText];
+    [self.remoteSelector addTarget:self action:@selector(remoteStreamingChanged) forControlEvents:UIControlEventValueChanged];
+}
+
+- (void) remoteStreamingChanged
+{
+    //This function can be used to reconfigure the settings view to offer more remote streaming options (i.e. reduce the audio frequency to 24kHz, enable/disable the HEVC bitrate multiplier, ...)
```
cgutman: Please keep a space between `//` and the start of the comment per the code style of the rest of the project.

Moonlight macOS/AppDelegate.m
```diff
@@ -0,0 +1,212 @@
+//
```
cgutman: Can we share some or all of this code with iOS?

Moonlight macOS/Input/Control.m
```diff
@@ -0,0 +1,188 @@
...
+#ifdef _MSC_VER
+#define snprintf _snprintf
+#endif
+
+static bool verbose = true;
+Controller* _controller;
+ControllerSupport* _controllerSupport;
+
+void onButtonDown(struct Gamepad_device * device, unsigned int buttonID, double timestamp, void * context) {
+    if (verbose) {
```
cgutman: We shouldn't depend on `verbose` being true to set button flags. There are similar bugs in `onButtonUp` and `onAxisMoved`

Moonlight macOS/Input/ControllerSupport.m
```diff
@@ -0,0 +1,63 @@
+//
```
cgutman: Can we share this with iOS?

Moonlight macOS/Input/Controller.h
```diff
@@ -0,0 +1,29 @@
+//
```
cgutman: This should be shared with iOS.

Moonlight macOS/Utility/keyboardTranslation.m
```diff
@@ -0,0 +1,163 @@
...
+//  Copyright © 2018 Felix Kratz. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "keyboardTranslation.h"
+#import <Limelight.h>
+
+CGKeyCode modifierKeyFromEvent(int keyModifier) {
+    switch (keyModifier) {
+        case 131330:
```
cgutman: Surely there are `const`s or `#define`s for these

Moonlight macOS/Utility/keyboardTranslation.m
```diff
@@ -0,0 +1,163 @@
...
+        //    return 0xA5; //RALT
+        case 262401:
+            return 0xA2; //LCTRL
+
+        default:
+            return 0x00;
+    }
+}
+
+char keyModifierFromEvent(int keyModifier)
```
cgutman: `keyModifierFromEvent` and `modifierKeyFromEvent` have names are easy to confuse with each other. Can you rename one or both?

Moonlight macOS/Moonlight_macOS.xcdatamodeld/Limelight 0.3.0.xcdatamodel/contents
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
cgutman: Can we avoid duplicating these data models from iOS?

Moonlight macOS/Utility/keepAlive.m
```diff
@@ -0,0 +1,30 @@
...
+
+#import <Foundation/Foundation.h>
+#import <IOKit/pwr_mgt/IOPMLib.h>
+#import "keepAlive.h"
+
+
+@implementation keepAlive
+
+CFStringRef reasonForActivity= CFSTR("Moonlight keeps the system awake");
+IOPMAssertionID assertionID = 1984201141;
```
cgutman: What's this magic number mean?

Moonlight macOS/Utility/keyboardTranslation.m
```diff
@@ -0,0 +1,163 @@
...
+
+#import <Foundation/Foundation.h>
+#import "keyboardTranslation.h"
+#import <Limelight.h>
+
+CGKeyCode modifierKeyFromEvent(int keyModifier) {
+    switch (keyModifier) {
+        case 131330:
+            return 0xA0; //LSHIFT
+            //case 131332:              TODO: This will lockup the modifiers
```
cgutman: What's the TODO mean exactly?

Moonlight macOS/ViewController/ViewController.m
```diff
@@ -0,0 +1,245 @@
...
+    _streamConfig.appID = [_sortedAppList[_popupButtonSelection.indexOfSelectedItem] id];
+    [self transitionToStreamView];
+}
+
+- (IBAction)textFieldAction:(id)sender {
+    [self buttonConnectPressed:self];
+}
+
+- (IBAction)buttonConnectPressed:(id)sender {
+    _host = _textFieldHost.stringValue;
```
cgutman: Can we share some of these functions in this class with iOS?

moonlight-common/moonlight-common_mac.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,510 @@
+// !$*UTF8*$!
```
cgutman: Is there a way to avoid having a duplicate moonlight-common project for Mac?

Moonlight macOS/ViewController/StreamFrameViewController.m
```diff
@@ -0,0 +1,123 @@
...
+
+-(ViewController*) _origin {
+    return _origin;
+}
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    [keepAlive keepSystemAlive];
+    self.streamConfig = _streamConfig;
+    Gamepad_detectDevices();
```
cgutman: It looks like `Gamepad_detectDevices()` should only be called after `Gamepad_init()`

Moonlight macOS/ViewController/StreamFrameViewController.m
```diff
@@ -0,0 +1,123 @@
...
+
+@interface StreamFrameViewController ()
+@end
+
+@implementation StreamFrameViewController {
+    StreamManager *_streamMan;
+    StreamConfiguration *_streamConfig;
+    NSTimer* _timer;
+    ViewController* _origin;
+    struct Gamepad_device* device;
```
cgutman: What uses this `device` variable?

Moonlight macOS/Input/Control.m
```diff
@@ -0,0 +1,188 @@
...
+            default:
+                break;
+        }
+        [_controllerSupport updateFinished:_controller];
+    }
+}
+
+void onDeviceAttached(struct Gamepad_device * device, void * context) {
+    if (verbose) {
+        printf("Device ID %u attached (vendor = 0x%X; product = 0x%X) with context %p\n", device->deviceID, device->vendorID, device->productID, context);
```
cgutman: You should create a `Controller` instance for each `struct Gamepad_device`, store it in an `NSMutableDictionary` and use that in the input event callbacks, rather than a single `Controller` object. This will give you proper multi-controller support. See iOS's ControllerSupport.m.

Moonlight macOS/Input/ControllerSupport.m
```diff
@@ -0,0 +1,63 @@
...
+}
+
+-(void) updateFinished:(Controller*)controller
+{
+    _controllerNumbers = 0;
+     _controllerNumbers |= (1 << 0);
+    controller.playerIndex = 0;
+    [_controllerStreamLock lock];
+    @synchronized(controller) {
+        LiSendMultiControllerEvent(controller.playerIndex, _controllerNumbers | 0, controller.lastButtonFlags, controller.lastLeftTrigger, controller.lastRightTrigger, controller.lastLeftStickX, controller.lastLeftStickY, controller.lastRightStickX, controller.lastRightStickY);
```
cgutman: `| 0` is a no-op

Moonlight macOS/Input/ControllerSupport.m
```diff
@@ -0,0 +1,63 @@
...
+-(void) clearButtonFlag:(Controller*)controller flags:(int)flags
+{
+    @synchronized(controller) {
+        controller.lastButtonFlags &= ~flags;
+    }
+}
+
+-(void) updateFinished:(Controller*)controller
+{
+    _controllerNumbers = 0;
```
cgutman: If you go for multi-controller support, I'd make `updateFinished` take the active controller bitmask as a parameter, since you won't have easy access to it here.

Limelight/Stream/Connection.m
FelixKratz: I was wrong here, sorry. I assumed that a connection via Internet would loose some frames here and there, but it doesn't seem so. I have now tested it with the Network Link Conditioner and came to the conclusion that it actually doesn't work.

Limelight/Stream/Connection.m
FelixKratz: My mac is relatively old and doesn't have hardware support, yet it looks much better than H264 on low bitrate connections. My idea would be to either let the user force HEVC or at least force it when the remote streaming optimizations are selected.

Moonlight macOS/Input/Control.m
FelixKratz: True. Fixed

Moonlight macOS/Input/ControllerSupport.m
FelixKratz: Done.

Moonlight macOS/Input/Controller.h
FelixKratz: Done.

Moonlight macOS/Utility/keyboardTranslation.m
FelixKratz: I have used the const's for now.

Moonlight macOS/Utility/keyboardTranslation.m
FelixKratz: I decided to rename both, maybe this could also be the foundation for a virtual keyboard on ios? I don't know if the keyCodes are the same though.

Moonlight macOS/Moonlight_macOS.xcdatamodeld/Limelight 0.3.0.xcdatamodel/contents
FelixKratz: Yes, fixed.

Moonlight macOS/Utility/keepAlive.m
FelixKratz: Actually, nothing at all. I cleaned it up.

Moonlight macOS/Utility/keyboardTranslation.m
FelixKratz: There seems to be a difference in the handling of keyUp-Events for the right side modifiers, this lead to a bug where the key got "stuck", i have now evaded this by using the keyModifier const's. It seems that there are no separate const's for the left and right side modifiers.

moonlight-common/moonlight-common_mac.xcodeproj/project.pbxproj
FelixKratz: About this one i really don't know exactly, i tried to use the same one but the linker complained about the architecture, maybe there is a way, but i don't know. The answer to this would possibly also help with the duplicate opus lib.

Moonlight macOS/ViewController/StreamFrameViewController.m
FelixKratz: This is a leftover from testing, deleted.

Moonlight macOS/AppDelegate.m
FelixKratz: Yes, done.

Moonlight macOS/Input/Control.m
FelixKratz: This is a good idea, i can implement that, but have no way of testing it because i only have one controller. Or is there a virtual controller tool for testing?

Moonlight macOS/ViewController/ViewController.m
FelixKratz: Yes, but not as smooth as in most of the other classes. We'd have to create a helper class that both view controllers can call. Or are there better ways?

Moonlight macOS/ViewController/StreamFrameViewController.m
FelixKratz: It seems like Gamepad_detectDevices() is not required at all for initializing the gamepad.
Even when connecting during streaming there is no problem at all. I will look into this when implementing multi controller support.

cgutman:
Sorry, I've been a bit busy this week. I'll try to look at this again by this weekend.

cgutman:
I'm going to make some additional changes on top of this, but it looks good to merge as is.

cgutman:
I refactored a bunch of the code in https://github.com/moonlight-stream/moonlight-ios/commit/e8832ed746cc829adbe6fb40d59f93599cb8486f

Let me know if I broke anything in the process.

cgutman:
You should also rebase your master branch on the master branch here before making any further changes. Otherwise merging will be very messy.

FelixKratz:
Everything works as before. I noticed a bug with the controller support that was also present in the previous build. When the controller recently got connected and a button is pressed, there will sometimes be a crash. If the controller is connected for a while this will not happen. I don't quite understand the bug. Any ideas?

PR #341, opened 2018-10-18
tvOS app UI/UX fixes

ascagnel:
- [x] do not quit streams when pressing "B" on MFi controllers
- [x] do not quit streams immediately on pressing the menu button on MFi controllers
- [x] fill out missing images (app icon, launch icon, etc)
- [x] make the selection indicator more pronounced
- [x] added a settings bundle

sorry if this is bad, this is my first time dipping my toes into iOS/tvOS development


cgutman:
Thanks, were you planning to add commits for all of the items you have listed prior to merge?

ascagnel:
Yup!

I have a quick & dirty change for the focus indicator (item 4), but it’s not great and doesn’t look/behave at all like the default tvOS focus UX. I also started on the menu button functionality, but that’ll be weird since it’s used for multiple things in this case (system UI navigation, in-stream pause, and backing out of a stream).

ascagnel:
i was able to get proper tvOS-style focused icons working (i'm still new to iOS/tvOS dev work), and went through and made some quick icons.

Limelight/UIAppView.m
```diff
@@ -64,10 +68,15 @@ - (void) updateAppImage {
     if ([_app.id isEqualToString:_app.host.currentGame]) {
         // Only create the app overlay if needed
         _appOverlay = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"Play"]];
+#if TARGET_OS_TV
+        _appOverlay.adjustsImageWhenAncestorFocused = YES;
+        _appOverlay.userInteractionEnabled = YES;
```
cgutman: The app overlay appears to be getting clipped now on the top and bottom corners of the triangle and also appears to have a strange glow around it which is also clipped when selected.

Limelight/ViewControllers/MainFrameViewController.m
```diff
@@ -1046,12 +1045,20 @@ - (void) enableNavigation {
     self.navigationController.navigationBar.topItem.rightBarButtonItem.enabled = YES;
 }
 
+#if TARGET_OS_TV
+- (BOOL)canBecomeFocused {
+    return YES;
```
cgutman: Just curious, what does this do?

Limelight/UIAppView.m
```diff
@@ -32,7 +32,11 @@ - (id) initWithApp:(TemporaryApp*)app cache:(NSCache*)cache andCallback:(id<AppC
         noImage = [UIImage imageNamed:@"NoAppImage"];
     }
     
+#if TARGET_OS_TV
+    _appButton = [UIButton buttonWithType:UIButtonTypeSystem];
```
cgutman: Do we get some sort of styling difference using `UIButtonTypeSystem`?

Limelight/UIAppView.m
```diff
@@ -91,8 +100,20 @@ - (void) updateAppImage {
...
+#if TARGET_OS_TV
+            //custom image to do TvOS hover popup effect
+            UIImageView *imageView = [[UIImageView alloc] initWithImage:appImage];
+            imageView.userInteractionEnabled = YES;
+            imageView.adjustsImageWhenAncestorFocused = YES;
+            imageView.frame = CGRectMake(0, 0, 200, 265);
+            [_appButton addSubview:imageView];
+            
+            _appButton.frame = CGRectMake(0, 0, 200, 265);
+            self.frame = CGRectMake(0, 0, 200, 265);
```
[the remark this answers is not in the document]
ascagnel: Fiddling around with what looked ok.  coming back to this again after a few days, I think it's actually a little too big compared to other "box art" views.

Limelight/ViewControllers/MainFrameViewController.m
ascagnel: tvOS has a "focus engine" that's used to move from item to item using the remote's trackpad or the left analog/d-pad of a connected controller.  Returning `YES` here sets a flag that the given element can be be focused by the focus engine.

A couple of links:
https://developer.apple.com/documentation/uikit/uiview/1622584-canbecomefocused?language=objc
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppleTV_PG/WorkingwiththeAppleTVRemote.html#//apple_ref/doc/uid/TP40015241-CH5-SW5

It's worth noting that I found I didn't *need* to override this function for it to work correctly, but per the API, it should still be necessary.

Limelight/UIAppView.m
ascagnel: Without `UIButtonTypeSystem`, I wasn't getting the tvOS style "zoomed in, rotates a little on touch" style selectors that are used throughout the system.

ascagnel:
quick update on this -- the big thing to fix is when you hit menu on an attached MFi controller.  StreamFrameViewController's `willMoveToParentViewController` is getting called when hitting menu, before the ControllerSupport's pause handler can be called.  If I move the stopStream & controller cleanup out of that method, I can hear the stream correctly handling the input, but the StreamFrameViewController has already been popped off the stack.  I'm looking into how the `willMoveToParentViewController` handler can be suppressed in this case, unless there's some other action that should trigger it.

Limelight/UIAppView.m
ascagnel: this should be resolved now -- the icon won't animate, but it looks better than a giant overlay icon

Moonlight TV/Base.lproj/Main.storyboard
```diff
@@ -39,7 +39,6 @@
                     <collectionView key="view" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="scaleToFill" dataMode="prototypes" prefetchingEnabled="YES" id="8P0-wP-bDe">
                         <rect key="frame" x="0.0" y="0.0" width="1920" height="1080"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" red="0.33333334329999997" green="0.33333334329999997" blue="0.33333334329999997" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
```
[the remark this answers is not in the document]
ascagnel: i fixed this by wrapping the white color assignments in `#if !TARGET_OS_TV` blocks

Limelight/UIAppView.m
```diff
@@ -91,8 +100,20 @@ - (void) updateAppImage {
         // TODO: Improve no-app image detection
         if (!(appImage.size.width == 130.f && appImage.size.height == 180.f) && // GFE 2.0
             !(appImage.size.width == 628.f && appImage.size.height == 888.f)) { // GFE 3.0
+#if TARGET_OS_TV
+            //custom image to do TvOS hover popup effect
```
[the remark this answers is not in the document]
ascagnel: this should be resolved now

ascagnel:
Also, this might just be my setup, but I'm getting about 200ms or so of latency between video and audio -- changes to video happen within a frame as best I can tell (ATV & host PC are both on wireline connections), but audio is noticeably delayed.

ascagnel:
Since I won’t be able to finish this PR for the foreseeable future, is it worth trying to get it merged as-is?

cgutman:
There are 2 main issues remaining:
- The play button is not properly placed on the box art (it's up in the upper left corner)
- There's no way to distinguish the selected icon on the PC selection view anymore

I think it's okay to merge right now. It definitely improves more than it breaks. It just needs some polish before release. I should have more time to work it in January.

ascagnel:
Some observations from the code I looked at:

- you’ll likely need to apply the same UIButton treatment to the PC select icons to have them appear correctly. UIButtons probably won’t handle transparent stuff very well, unless you can do some fakery with alpha channels
- I wanted to re-work the play button & text (really, the entire draw of buttons) to layer everything on like I’m doing for text on the default icon — it should play nicely with the play icons (no random gloss effects or static arrows)

PR #351, opened 2018-12-06
more appletv tweaks

ascagnel:
- add a green "now playing" glow to the currently-running app
- fix a bug for when you want to use a controller directly with the host

### DIFF
--- a/Limelight/UIAppView.m
+++ b/Limelight/UIAppView.m
@@ -76,19 +76,24 @@ static UIImage* noImage;
     self.frame = CGRectMake(0, 0, noImage.size.width, noImage.size.height);
     
     if ([_app.id isEqualToString:_app.host.currentGame]) {
+#if TARGET_OS_TV
+        _appButton.layer.masksToBounds = NO;
+
+        _appButton.layer.shadowColor = [UIColor greenColor].CGColor;
+        _appButton.layer.shadowOffset = CGSizeMake(5,8);
+        _appButton.layer.shadowOpacity = 0.7;
+#else
         // Only create the app overlay if needed
         _appOverlay = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"Play"]];
-#if !TARGET_OS_TV
         _appOverlay.layer.shadowColor = [UIColor blackColor].CGColor;
         _appOverlay.layer.shadowOffset = CGSizeMake(0, 0);
         _appOverlay.layer.shadowOpacity = 1;
         _appOverlay.layer.shadowRadius = 2.0;
-#endif
-        
         [self addSubview:_appOverlay];
         
         _appOverlay.frame = CGRectMake(0, 0, noImage.size.width / 2.f, noImage.size.height / 4.f);
         [_appOverlay setCenter:CGPointMake(self.frame.size.width/2, self.frame.size.height/6)];
+#endif
     }
     
     BOOL noAppImage = false;
@@ -120,10 +125,11 @@ static UIImage* noImage;
 #else
             _appButton.frame = CGRectMake(0, 0, appImage.size.width / 2, appImage.size.height / 2);
             self.frame = CGRectMake(0, 0, appImage.size.width / 2, appImage.size.height / 2);
-#endif
+
             _appOverlay.frame = CGRectMake(0, 0, self.frame.size.width / 2.f, self.frame.size.height / 4.f);
             _appOverlay.layer.shadowRadius = 4.0;
             [_appOverlay setCenter:CGPointMake(self.frame.size.width/2, self.frame.size.height/6)];
+#endif
             [_appButton setBackgroundImage:appImage forState:UIControlStateNormal];
             [self setNeedsDisplay];
         } else {
@@ -176,7 +182,9 @@ static UIImage* noImage;
     // Update the app image if neccessary
     if ((_appOverlay != nil && ![_app.id isEqualToString:_app.host.currentGame]) ||
         (_appOverlay == nil && [_app.id isEqualToString:_app.host.currentGame])) {
+#if !TARGET_OS_TV
         [self updateAppImage];
+#endif
     }
     
     // Stop updating when we detach from our parent view
--- a/Limelight/ViewControllers/MainFrameViewController.m
+++ b/Limelight/ViewControllers/MainFrameViewController.m
@@ -477,7 +477,11 @@ static NSMutableSet* hostList;
     
     // multiController must be set before calling getConnectedGamepadMask
     _streamConfig.multiController = streamSettings.multiController;
-    _streamConfig.gamepadMask = [ControllerSupport getConnectedGamepadMask:_streamConfig];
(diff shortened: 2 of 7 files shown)


Limelight/UIAppView.m
```diff
@@ -108,6 +110,19 @@ - (void) updateAppImage {
             imageView.userInteractionEnabled = YES;
             imageView.adjustsImageWhenAncestorFocused = YES;
             imageView.frame = CGRectMake(0, 0, 200, 265);
+
+            if (isCurrentGame) {
+                UIImageView *playIcon = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"Play"]];
```
cgutman: I don't really like the way our play button scales up to tvOS (and frankly it's not great on iOS either). It covers most of the app view and the contrast is not good against some box art, so it's easy to lose the play button unless you're really looking for it.

Maybe we should try doing something else to indicate the running app (maybe a green border highlight or something?).

Limelight/UIComputerView.m
```diff
@@ -30,6 +30,15 @@ - (id) init {
     _hostButton = [UIButton buttonWithType:UIButtonTypeCustom];
     [_hostButton setContentEdgeInsets:UIEdgeInsetsMake(0, 4, 0, 4)];
     [_hostButton setBackgroundImage:[UIImage imageNamed:@"Computer"] forState:UIControlStateNormal];
+    
+#if TARGET_OS_TV
+    UIImageView *backgroundIcon = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"Computer"]];
+    backgroundIcon.userInteractionEnabled = YES;
+    backgroundIcon.adjustsImageWhenAncestorFocused = YES;
+    backgroundIcon.clipsToBounds = NO;
+    [_hostButton addSubview:backgroundIcon];
```
cgutman: You can still see the original size host icon if you look closely during the animation. Can we animate this without need a separate button and image view?

Limelight/ViewControllers/MainFrameViewController.m
```diff
@@ -472,7 +472,11 @@ - (void) prepareToStreamApp:(TemporaryApp *)app {
     
     // multiController must be set before calling getConnectedGamepadMask
     _streamConfig.multiController = streamSettings.multiController;
-    _streamConfig.gamepadMask = [ControllerSupport getConnectedGamepadMask:_streamConfig];
+    if (!streamSettings.disableTvOSController) {
```
cgutman: Why are you thinking of a separate option for this? TvOS doesn't have on-screen controls, so we should report 0 gamepads unless a real one is connected to the Apple TV

ascagnel: Even if no controllers are connected, the connected remote acts as a controller.  I'll see if I can work around that, but it may not be ideal.

cgutman: Yeah, we should exclude micro gamepads (which I thought we did already).

ascagnel: i figured what the source of this is -- if no controllers are connected and multi-controller support is disabled, the controller mask will always be set to `0x1`, regardless of the on-screen control setting.  i updated this so that, if running on tvOS, the mask will not be updated (and default to 0).  this also means the micro-controller (siri remote) can also act as a mouse, even with a controller connected to the pc.

i was also able to remove the new setting.

Limelight/UIComputerView.m
ascagnel: this change has been reverted

Limelight/UIAppView.m
ascagnel: i added a green border highlight similar to what's now on the host selection view

ascagnel:
added a small tweak -- do not show a "connection terminated" message when it's from an app quitting successfully (`errorCode` of 0).

Limelight/ViewControllers/StreamFrameViewController.m
```diff
@@ -191,12 +191,15 @@ - (void)connectionTerminated:(long)errorCode {
     dispatch_async(dispatch_get_main_queue(), ^{
         // Allow the display to go to sleep now
         [UIApplication sharedApplication].idleTimerDisabled = NO;
-        
-        UIAlertController* conTermAlert = [UIAlertController alertControllerWithTitle:@"Connection Terminated" message:@"The connection was terminated" preferredStyle:UIAlertControllerStyleAlert];
-        [conTermAlert addAction:[UIAlertAction actionWithTitle:@"OK" style:UIAlertActionStyleDefault handler:^(UIAlertAction* action){
+        if (errorCode == 0) {
```
cgutman: This actually wasn't accurate because moonlight-common-c wasn't propagating the error code correctly. I think it was always 0.

But I liked the idea, so I reverse engineered the server termination message that sends us the termination reason and now pass 0 to this callback if the server termination code is SERVER_TERMINATED_INTENDED.

ascagnel:
rebased this on top of master; LGTM?

cgutman:
It looks mostly good. I don't understand why there's divergence between the behavior of multi-controller mode on tvOS and on iOS. It should default to enabled rather than disabled, then you don't need to hack the `mask = 0x1;` line because controller hot-plug is enabled by default.

If you're okay with me making that change, I can merge it and do the change. If not, we can discuss further or you can make the change yourself prior to merging.

ascagnel:
I'm fine to revert the mask change; however, I did see it happening when I was attempting to use a controller paired with the host PC while streaming to my phone, so maybe it's larger than just on the TV app?

cgutman:
If you're talking about gamepad 0 (player 1) being always present when multi-controller is set to off, that is the expected behavior of that configuration.

In the default "auto" position:
- Gamepads are connected to the host PC only when physically connected
- Gamepads are disconnected from the host PC when disconnected from the client device
- Gamepads are assigned unique controller numbers

In the "off" position:
- Gamepad 0 is always connected to the host PC, even if a gamepad is detached or not yet attached to the client
- Gamepad 0 is never disconnected from the host PC
- All physically connected gamepads send input as gamepad 0 (player 1)

The non-default "off" position is only intended for solving issues with games that exhibit specific controller hotplugging issues. In Left 4 Dead, controller detection happens only at game startup and disconnected controllers are never re-detected after reconnection. That means if you disconnect or switch controllers on the client with multi-controller set to "auto", you will forever lose gamepad control until the game is restarted. Likewise, if you first start the game then connect your gamepad (instead of the other way around) with multi-controller set to "auto", you will not be able to control the game until it's restarted with the gamepad attached.

Are you seeing behavior consistent with this or something different?

ascagnel:
This is consistent with what I'm seeing.  Let me test to see if the masking works with multi-controller turned on without the mask fix on the TV app.

ascagnel:
After some testing, defaulting `multiController` on and pairing a controller before the stream begins works as described above; I removed the tvOS exception to the mask.

cgutman:
Looks great, thanks.

PR #426, opened 2020-10-27
Bump max bitrate to 150Mbps to line up with android and qt version

nauxliu:
Refs:
https://github.com/moonlight-stream/moonlight-android/commit/ca0aee58ab4250f82f60a892a258070c7f7cbcb2
https://github.com/moonlight-stream/moonlight-qt/blob/3e36357315937d5083f84a1be76012622628e30d/app/gui/SettingsView.qml#L325


cgutman:
Thanks

Issue #442, opened 2020-12-24
Requesting more relevant iOS resolutions


brianbauer42:
I would really appreciate the addition of different resolutions. Either a custom resolution, or perhaps the ability to detect the resolution of the device the app is running on. The jump from 1080p to 4k is pretty large, and while they are common resolutions on televisions, they are not used on any iPads or iPhones. 16:9 is also not the correct aspect ratio for any iPad or iPhone.

I was thrilled to see how well Disco Elysium played with touch on a 2018 iPad Pro 11", but the 16:9 aspect ratio feels very restricted and there's a lot of wasted screen. On a handheld size device, it's important to use every pixel you've got, so I set a custom resolution in nvidia control panel to the native resolution of my iPad, launched Disco Elysium and was able to set it to the same - 2388 x 1668. 

Awesome!

But wait... Moonlight will only stream 1080p or 4k :( Sad times.

Otherwise, this project is awesome! I find the streaming to be much smoother than with Steam Link.

bakedChicken:
Hello. I was able to make Moonlight for iOS app run stream in iPad 11" 2020 native (2388x1668) resolution.
As you mentioned I also have to change my monitor resolution to 2388x1668.
You can see how GTA IV is running with these settings:

![C2EC868F-9B84-4938-9291-0A3AEC750CB2_1_105_c](https://user-images.githubusercontent.com/22817721/107450137-88269c00-6b66-11eb-8117-3c4b34313430.jpeg)
![98C97644-455D-4920-94B0-E1A5480A5ED7_1_105_c](https://user-images.githubusercontent.com/22817721/107450153-8fe64080-6b66-11eb-9a31-1f4d994f98b2.jpeg)

Clearly there is issue, I believe no game is designed to be run in 199/139 aspect ratio.
Do you think is it worth to play game like this?
Thanks.

cgutman:
Custom resolutions are supported in v8.0.0

PR #454, opened 2021-02-25
Added support for native stream resolution

bakedChicken:
Added support for native stream resolution. More information about result could be found in this issue: https://github.com/moonlight-stream/moonlight-ios/issues/442

Of course this is not for every game, but I found that Civilization 6 supports this resolution well.

### DIFF
--- a/Limelight/ViewControllers/SettingsViewController.h
+++ b/Limelight/ViewControllers/SettingsViewController.h
@@ -23,6 +23,7 @@
 @property (strong, nonatomic) IBOutlet UISegmentedControl *btMouseSelector;
 @property (strong, nonatomic) IBOutlet UISegmentedControl *statsOverlaySelector;
 @property (strong, nonatomic) IBOutlet UIScrollView *scrollView;
+@property (strong, nonatomic) IBOutlet UILabel *resolutionLabel;
 
 #pragma clang diagnostic push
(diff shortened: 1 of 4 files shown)


cgutman:
I'd like the native resolution to appear in the button label, so users don't have to go look up what their device's native resolution is. Can you implement that?

bakedChicken:
You mean to show it just as an additional information in settings or actually extract this option out of segments and implementing this as a button with actual resolution as a button label?

cgutman:
I think it would be nice for the button itself to have the actual resolution. You can either insert the segment at runtime, or just change the title at runtime.

bakedChicken:
I have added a label with chosen resolution information. It looks like this:

7C452E0A-7440-4787-96B7-6DA181BB5F85_1_105_c

cgutman:
Hmm, what I had in mind was the segment button itself displaying the native resolution. But at the very least, we need to get the segment sizing correct so the text isn't cut off.

nkulung:
A resolution option (either native or somewhere between 1080P and 4K) would be great to have available. Is there a dropdown element that could be used in place of the segments? 

I think a 2K resolution would be perfect, A 4:3 aspect ratio resolution would be great as well, but I figure that it doesn't scale that well with certain applications/games.

cgutman:
Superseded by #476

PR #476, opened 2021-10-12
Added support for native stream resolution (ui fixed)

hbouhadji:
Fixed ui version of #454 (by @bakedChicken)

+ Keeps the safe area (for devices with the notch and rounded edges etc...)

Improvements to-do:
- maybe add a select above the resolution to ask if the user want to keep the safe area or if he want the stream to be full full screen 
- or/and a selector to ask the aspect ratio 4/3, 16/9, 18/9 (iphone with safe area), 19.5/9 (full screen)
- wanted to use popup button to replace all the segmented controls but it's available only since iOS 14. An alternative can be action sheets with a custom resolution option
- calculate the safe area for device < iOS 11

⚠️ Warning:
For some reasons when I'm creating and switching to a custom resolution (with special aspect ratio) on the nvidia control panel, the cursor and the click of the mouse are not aligned correctly. I lost hours trying to debug that but I found a "solution". You have to create the custom resolution on the nvidia control panel but if you want to use it, you have to switch the resolution from the windows panel (exactly like this -> https://www.youtube.com/watch?v=3KoqOFG7jsU) and it will work perfectly. I think that's worth a comment on the wiki/readme.

(Tested on iPhone 13 Pro Max and iPad Pro 11")
some screens:
IMG_5828
IMG_5837
IMG_5832


thomasaull:
@cgutman Is there any plans when this will get merged and released? I’m kind of waiting for this feature 😬

TizianoMattei:
this feature is a game changer for lots of different workflows. really hoping it gets merged.

cgutman:
Yeah, I would like to merge it. The problem is the issue noted in the PR description (which is not specific to this particular native resolution PR).

Apple's decoder takes into account the pixel aspect ratio encoded in the H.264/HEVC stream to perform scaling that prevents our StreamView from knowing where the user input should map to in touch mode. We need to either find a way to opt out of that or strip the pixel aspect ratio from the H.264 and HEVC SPS data.

TizianoMattei:
> Yeah, I would like to merge it. The problem is the issue noted in the PR description (which is not specific to this particular native resolution PR).
> 
> Apple's decoder takes into account the pixel aspect ratio encoded in the H.264/HEVC stream to perform scaling that prevents our StreamView from knowing where the user input should map to in touch mode. We need to either find a way to opt out of that or strip the pixel aspect ratio from the H.264 and HEVC SPS data.

Thanks for the update and clarification on the matter!
I see, that makes sense, didnt consider it as i rarely use touch mode. Would it be possible to run a beta version in the meantime on testflight?

thomasaull:
@cgutman Would it be an option to disable touch mode when using the custom resolution until there is found a better solution?

hbouhadji:
@cgutman 
I had new needs (ability to change aspect ratio) so I have an alert with two textfield to ask the width and the height.

IMG_6039
IMG_6038

I didn't had the bug with the cursor and the click of the mouse not aligned correctly.

cgutman:
Rebased and merged. Thanks!

PR #477, opened 2021-10-16
HDR UI setting for iOS

jgreer013:
Added UI setting for HDR and passthrough said setting.

Unfortunately, even when working with an HDR enabled display, the device isn't displaying the image properly. I've been testing with Destiny 2 on Steam and it looks washed out no matter what.

### DIFF
--- a/Limelight/ViewControllers/SettingsViewController.m
+++ b/Limelight/ViewControllers/SettingsViewController.m
@@ -163,6 +163,16 @@ static const int bitrateTable[] = {
         [self.framerateSelector removeSegmentAtIndex:2 animated:NO];
     }
     
+    bool enableHdr = false;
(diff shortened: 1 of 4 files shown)


Limelight/ViewControllers/SettingsViewController.m
```diff
@@ -314,7 +331,7 @@ - (void) saveSettings {
                      multiController:multiController
                            audioOnPC:audioOnPC
                              useHevc:useHevc
-                           enableHdr:NO
+                           enableHdr:enableHDR
```
shokakucarrier, line 334: use YES for bool in Objective-C

cgutman:
There were a bunch of merge conflicts due to the frame pacing changes, so I just redid this myself.

Fixed by 54d1851c7828a887346a7493a1eb25dc801ea82e

PR #482, opened 2021-12-20
Add Vsync feature

### LINKED ISSUE #427: Occasional micro stuttering
iOS, iPadOS, TVOS while moonlight is running there are micro stuttering, maybe there should be one more frame buffer when streaming...

### LINKED ISSUE #478: tvOS settings not saving & stuttering stream
**Describe the bug**
Whenever I go and change settings in the tvOS moonlight app, if I then quit the app, the settings are restored back to default. I am also experiencing problems with the stream itself, about every second the stream freezes audio and video for a quick time. I'm not sure if these problems are related.

**Specifications**
Apple TV 4K over WiFi 6 (cable doesn't appear to help)
PC with RTX 2070 over ethernet on same LAN as the Apple TV.
GeForce Experience version: 3.24.0.123
Driver version: 496.76
OS: Windows 11 21H2 (22000.348)
LG TV: 4K, 50 hz with HDR

**Steps to reproduce**
With quit I refer to exiting the app fully by sliding it off.

1. Open Moonlight tvOS app
2. Click Settings
3. Change settings
4. Play a game then quit the app (or quit right away, doesn't matter) (It seems that if you play a game right away without quitting the app, it will keep the settings until you quit).
5. Go and check settings
6. Settings have been restored for some reason

**Screenshots**
IMG_1113

felipejfc:
Adds a vsync feature and setting toggle. This should fix #427, partially fix #478 and fix many others reports of micro stuttering in iOS devices. The cost is obviously a little added latency, however, I made it configurable if you want it on or off.

The way it works is right after receiving the frame, instead of sending to AVSampleBufferDisplayLayer for immediate decoding and rendering, we wait for a sync from displayLinkCallback which will happen in the same interval set by the refresh rate setting.

For many games, this makes the streaming experience much more enjoyable, the way it should be. Please consider merging @cgutman 

best!

### DIFF
--- a/Limelight/Limelight.xcdatamodeld/.xccurrentversion
+++ b/Limelight/Limelight.xcdatamodeld/.xccurrentversion
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>Moonlight v1.6.xcdatamodel</string>
+	<string>Moonlight v1.7.xcdatamodel</string>
 </dict>
 </plist>
--- a/Moonlight.xcodeproj/project.pbxproj
+++ b/Moonlight.xcodeproj/project.pbxproj
@@ -163,6 +163,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		566E9D2B2770B23A00EF7BFE /* Moonlight v1.7.xcdatamodel */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcdatamodel; path = "Moonlight v1.7.xcdatamodel"; sourceTree = "<group>"; };
 		693B3A9A218638CD00982F7B /* Settings.bundle */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.plug-in"; path = Settings.bundle; sourceTree = "<group>"; };
 		9803CCAB254F9EAF00EE185E /* ConnectionCallbacks.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = ConnectionCallbacks.h; sourceTree = "<group>"; };
 		98132E8C20BC9A62007A053F /* Moonlight v1.1.xcdatamodel */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcdatamodel; path = "Moonlight v1.1.xcdatamodel"; sourceTree = "<group>"; };
@@ -1428,6 +1429,7 @@
 		FB290D0519B2C406004C83CF /* Limelight.xcdatamodeld */ = {
 			isa = XCVersionGroup;
 			children = (
+				566E9D2B2770B23A00EF7BFE /* Moonlight v1.7.xcdatamodel */,
 				9819CC25254F2734008A7C8E /* Moonlight v1.6.xcdatamodel */,
 				98783FEA242EAC5D00F00EF4 /* Moonlight v1.5.xcdatamodel */,
 				98608BDD22DC0C2C000E5672 /* Moonlight v1.4.xcdatamodel */,
@@ -1440,7 +1442,7 @@
 				FB4678F21A51BDCB00377732 /* Limelight 0.3.0.xcdatamodel */,
 				FB290D0619B2C406004C83CF /* Limelight.xcdatamodel */,
 			);
-			currentVersion = 9819CC25254F2734008A7C8E /* Moonlight v1.6.xcdatamodel */;
+			currentVersion = 566E9D2B2770B23A00EF7BFE /* Moonlight v1.7.xcdatamodel */;
 			path = Limelight.xcdatamodeld;
 			sourceTree = "<group>";
 			versionGroupType = wrapper.xcdatamodel;
--- a/Limelight/Database/DataManager.h
+++ b/Limelight/Database/DataManager.h
@@ -22,6 +22,7 @@
                  multiController:(BOOL)multiController
                        audioOnPC:(BOOL)audioOnPC
                          useHevc:(BOOL)useHevc
+                  useFramePacing:(BOOL)useFramePacing
                        enableHdr:(BOOL)enableHdr
                   btMouseSupport:(BOOL)btMouseSupport
                absoluteTouchMode:(BOOL)absoluteTouchMode
--- a/Limelight/Database/DataManager.m
+++ b/Limelight/Database/DataManager.m
@@ -61,6 +61,7 @@
                  multiController:(BOOL)multiController
                        audioOnPC:(BOOL)audioOnPC
                          useHevc:(BOOL)useHevc
+                  useFramePacing:(BOOL)useFramePacing
                        enableHdr:(BOOL)enableHdr
                   btMouseSupport:(BOOL)btMouseSupport
                absoluteTouchMode:(BOOL)absoluteTouchMode
@@ -77,6 +78,7 @@
         settingsToSave.multiController = multiController;
         settingsToSave.playAudioOnPC = audioOnPC;
         settingsToSave.useHevc = useHevc;
+        settingsToSave.useFramePacing = useFramePacing;
         settingsToSave.enableHdr = enableHdr;
         settingsToSave.btMouseSupport = btMouseSupport;
         settingsToSave.absoluteTouchMode = absoluteTouchMode;
--- a/Limelight/Database/TemporarySettings.h
+++ b/Limelight/Database/TemporarySettings.h
@@ -19,6 +19,7 @@
 @property (nonatomic, retain) NSNumber * onscreenControls;
 @property (nonatomic, retain) NSString * uniqueId;
 @property (nonatomic) BOOL useHevc;
+@property (nonatomic) BOOL useFramePacing;
 @property (nonatomic) BOOL multiController;
 @property (nonatomic) BOOL playAudioOnPC;
 @property (nonatomic) BOOL optimizeGames;
--- a/Limelight/Database/TemporarySettings.m
+++ b/Limelight/Database/TemporarySettings.m
@@ -35,6 +35,7 @@
     self.framerate = [NSNumber numberWithInteger:[[NSUserDefaults standardUserDefaults] integerForKey:@"framerate"]];
     assert([self.framerate intValue] != 0);
     self.useHevc = [[NSUserDefaults standardUserDefaults] boolForKey:@"useHevc"];
+    self.useFramePacing = [[NSUserDefaults standardUserDefaults] boolForKey:@"useFramePacing"];
     self.playAudioOnPC = [[NSUserDefaults standardUserDefaults] boolForKey:@"audioOnPC"];
     self.enableHdr = [[NSUserDefaults standardUserDefaults] boolForKey:@"enableHdr"];
     self.optimizeGames = [[NSUserDefaults standardUserDefaults] boolForKey:@"optimizeGames"];
@@ -66,6 +67,7 @@
     self.height = settings.height;
     self.width = settings.width;
     self.useHevc = settings.useHevc;
+    self.useFramePacing = settings.useFramePacing;
     self.playAudioOnPC = settings.playAudioOnPC;
     self.enableHdr = settings.enableHdr;
     self.optimizeGames = settings.optimizeGames;
--- a/Limelight/Limelight.xcdatamodeld/Moonlight v1.7.xcdatamodel/contents
+++ b/Limelight/Limelight.xcdatamodeld/Moonlight v1.7.xcdatamodel/contents
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="19574" systemVersion="20G224" minimumToolsVersion="Xcode 7.3" sourceLanguage="Objective-C" userDefinedModelVersionIdentifier="">
+    <entity name="App" representedClassName="App" syncable="YES" codeGenerationType="class">
+        <attribute name="hdrSupported" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="hidden" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="id" attributeType="String" syncable="YES"/>
+        <attribute name="name" attributeType="String" syncable="YES"/>
+        <relationship name="host" maxCount="1" deletionRule="Nullify" destinationEntity="Host" inverseName="appList" inverseEntity="Host" syncable="YES"/>
+    </entity>
+    <entity name="Host" representedClassName="Host" syncable="YES" codeGenerationType="class">
+        <attribute name="address" optional="YES" attributeType="String" syncable="YES"/>
+        <attribute name="externalAddress" optional="YES" attributeType="String" syncable="YES"/>
+        <attribute name="ipv6Address" optional="YES" attributeType="String" syncable="YES"/>
+        <attribute name="localAddress" optional="YES" attributeType="String" syncable="YES"/>
+        <attribute name="mac" optional="YES" attributeType="String" syncable="YES"/>
+        <attribute name="name" attributeType="String" syncable="YES"/>
+        <attribute name="pairState" optional="YES" attributeType="Integer 32" defaultValueString="0" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="serverCert" optional="YES" attributeType="Binary" syncable="YES"/>
+        <attribute name="serverCodecModeSupport" attributeType="Integer 32" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="uuid" optional="YES" attributeType="String" syncable="YES"/>
+        <relationship name="appList" optional="YES" toMany="YES" deletionRule="Cascade" destinationEntity="App" inverseName="host" inverseEntity="App" syncable="YES"/>
+    </entity>
+    <entity name="Settings" representedClassName="Settings" syncable="YES" codeGenerationType="class">
+        <attribute name="absoluteTouchMode" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="bitrate" attributeType="Integer 32" defaultValueString="10000" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="btMouseSupport" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="enableHdr" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="framerate" attributeType="Integer 32" defaultValueString="60" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="height" attributeType="Integer 32" defaultValueString="720" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="multiController" attributeType="Boolean" defaultValueString="YES" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="onscreenControls" attributeType="Integer 32" defaultValueString="1" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="optimizeGames" attributeType="Boolean" defaultValueString="YES" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="playAudioOnPC" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="statsOverlay" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="uniqueId" attributeType="String" syncable="YES"/>
+        <attribute name="useFramePacing" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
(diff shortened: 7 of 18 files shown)


doctorpangloss:
I've explored this issue on the iPhone 12. You can see my work here - [git@github.com:doctorpangloss/moonlight-ios.git](https://github.com/doctorpangloss/moonlight-ios) - your approach doesn't work, because there are just more delivered frames than there are rendering frames on the 12.

felipejfc:
Did you actually try it? I have no problems in either my iPhone 12 or Apple TV and btw the way the code is right now it’s already drawing multiple frames in the same loop run if the queue count is >1 and since I set displayImmediately = true this is basically blending two frames.

doctorpangloss:
> Did you actually try it? 

Yes.

> I have no problems in either my iPhone 12 

Yes, the issue persists. Toggling `Vsync` on and off has no effect.

> I set displayImmediately = true this is basically blending two frames.

This doesn't blend frames. It skips a frame.

doctorpangloss:
I tried it. I don't know. I see fewer microstutters. But they're not eliminated. This stuff is a red herring. It has something to do with the `vsync` lock. But `displayImmediate` just skips frames.

felipejfc:
I would check refresh rate set in your monitor/game and your network performance. I’m currently traveling and I’ve been playing with this solution under 80ms ping conditions with 1080p and 20mbps bitrate and the games are very fluid right now.
In my house I’m able to do 4K 100mbps bitrate no problem without microstuttering.
Be sure to test if you’re actually getting 60fps in the pc you’re running the game and remember that only the fact you’re streaming if the setting is very high can cause a 15% overhead in your GPU. This is why I believe the best way to test if the stutter was gone is to try a lower resolution and bitrate first, so that you’re confident you’re not being limited by your own networking solution or in game settings being too demanding in your GPU

doctorpangloss:
> Try testing on a less demanding game, such as ori and the will of whisps, limit the in game fps to 60 and see how it goes, star with 1080p and keep going up

I only test with https://www.vsynctester.com on a desktop with a 3090.

> so I will be rendering always the last frame that arrived and will not lag behind.

If there are two picture frames per vsync, this will render as a frame skip.

I tried blending the two pictures per vsync here - https://github.com/doctorpangloss/moonlight-ios/blob/3dbe86eaec3c096352a0e4876e6061257ea43f6d/Limelight/Stream/VideoDecoderRenderer.m#L172 - but I get a magenta cast in the blended image because I'm just not super familiar with `vImage`. If you are experienced and know how to use vImage, sorting out why this produces a magenta cast will fix the issue.

felipejfc:
Cool, I will test with the sites you sent! Thanks 

Regarding the frameskip when >1 frames arrive in the same sync, I’m aware but I think it happens too little times for us to bother, I will do a test to validate the hypothesis but under a stable environment this should happen only during some occasional networking or GPU hiccups

doctorpangloss:
> Cool, I will test with the sites you sent! Thanks
> 
> Regarding the frameskip when >1 frames arrive in the same sync, I’m aware but I think it happens too little times for us to bother, I will do a test to validate the hypothesis but under a stable environment this should happen only during some occasional networking or GPU hiccups
> 
> === EDIT
> 
> So I tested with https://www.vsynctester.com and I got stable gray vsync with my fork and it turned on, as I said before it's impossible to be 100% stable since networking hiccups every now and then but it's very noticeable the difference when setting vsync on/off.
> 
> I will next see how many occurrences of >1 frame arriving in the same sync loop happens to see if it's worth blending images together

try the patch I said in the review - replace the contents of this block VideoDecoderRenderer.m 413-415 with
```
[NSThread sleepForTimeInterval:0.001f];
```
and observe the same apparent improvement in microstuttering. What I'm trying to say is that the vsync is a red herring, you're doing something else that's still interesting but it isn't the vsync. I'm not sure 100% what it is yet. Because it's not really vsync, you are still dispatch onto an async queue to enqueue the sample. Also, the name of the method is all we got, `enqueueSampleBuffer` which as it implies isn't guaranteed to be synchronous drawing. It doesn't say anywhere that it is. I don't know how it draws (immediate mode? deferred?).

When I tried to implement vsync, i.e. CADisplayLink callback is where the buffer gets enqueued, I didn't notice a significant difference. Also, I blinded myself for my tests (like a randomized control trial) and maybe you should too. Take a close look at my code, I've investigated this stuff a lot.

doctorpangloss:
Anyway if you or someone else you know could just fix the magenta cast in my vImage usage we will be done and this is fixed for good :)

felipejfc:
The approach I used here is the exact same as moonlight-qt (here)[https://github.com/moonlight-stream/moonlight-qt/blob/master/app/streaming/video/ffmpeg-renderers/vt.mm#L323-L331]

I think I understand what you say when you say this is not exactly vsync but I also believe true 100% vsync is impossible in this case as we are susceptible to networking oscillations. We don't know if by the time the next vsync happen I will have received already another frame from the network layer.

regardless I will try your branch, I just think there may be some overengineering there but let me take a look... One problem I already saw, you're using the wrong colortype and also you're not taking into consideration HDR

doctorpangloss:
> The approach I used here is the exact same as moonlight-qt [here](https://github.com/moonlight-stream/moonlight-qt/blob/master/app/streaming/video/ffmpeg-renderers/vt.mm#L323-L331)
> 
> I think I understand what you say when you say this is not exactly vsync but I also believe true 100% vsync is impossible in this case as we are susceptible to networking oscillations. We don't know if by the time the next vsync happen I will have received already another frame from the network layer.
> 
> regardless I will try your branch, I just think there may be some overengineering there but let me take a look... One problem I already saw, you're using the wrong colortype and also you're not taking HDR into consideration, it will break for sure

this is really helpful

> The approach I used here is the exact same as moonlight-qt here

yeah, but the difference is a macOS device correctly has a true 60.000 hz refresh rate, whereas the iPhone 12 does not. My bet is if you remove the lock, you won't notice much.

Limelight/Limelight.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>Moonlight v1.6.xcdatamodel</string>
+	<string>Moonlight v1. 7.xcdatamodel</string>
```
travis134, line 6: Looks like an extra space snuck in here.

Moonlight.xcodeproj/project.pbxproj
```diff
@@ -1424,7 +1442,7 @@
 				FB4678F21A51BDCB00377732 /* Limelight 0.3.0.xcdatamodel */,
 				FB290D0619B2C406004C83CF /* Limelight.xcdatamodel */,
 			);
-			currentVersion = 9819CC25254F2734008A7C8E /* Moonlight v1.6.xcdatamodel */;
+			currentVersion = 566E9D2B2770B23A00EF7BFE /* Moonlight v1. 7.xcdatamodel */;
```
travis134, line 1445: Same here, an extra space.

Moonlight.xcodeproj/project.pbxproj
```diff
@@ -1412,6 +1429,7 @@
 		FB290D0519B2C406004C83CF /* Limelight.xcdatamodeld */ = {
 			isa = XCVersionGroup;
 			children = (
+				566E9D2B2770B23A00EF7BFE /* Moonlight v1. 7.xcdatamodel */,
```
travis134, line 1432: Extra space.

travis134 (commented):
Looks like your find and replace for the version string included an extra space. I’d approve if I knew more about this code base, but I’m just an onlooker anxious to see this change get merged in case it can help fix the weird Apple TV frame rate judder problems with the Moonlight app.

Thanks for putting this fix together!

felipejfc:
> > The approach I used here is the exact same as moonlight-qt [here](https://github.com/moonlight-stream/moonlight-qt/blob/master/app/streaming/video/ffmpeg-renderers/vt.mm#L323-L331)
> > I think I understand what you say when you say this is not exactly vsync but I also believe true 100% vsync is impossible in this case as we are susceptible to networking oscillations. We don't know if by the time the next vsync happen I will have received already another frame from the network layer.
> > regardless I will try your branch, I just think there may be some overengineering there but let me take a look... One problem I already saw, you're using the wrong colortype and also you're not taking HDR into consideration, it will break for sure
> 
> this is really helpful
> 
> > The approach I used here is the exact same as moonlight-qt here
> 
> yeah, but the difference is a macOS device correctly has a true 60.000 hz refresh rate, whereas the iPhone 12 does not. My bet is if you remove the lock, you won't notice much.

Some notes

1) Your code has some huge memory leak, to trigger it run with 4k resolution and 100mbps bitrate
2) What you're doing is essentially the same I'm doing, both of us are rendering the frames during each CALinkCallback, the only difference is that  you tried to blend images together when you receive more than one in a loop and I skip one frame when this happens.

Honestly the approach you're proposing seems to spend more resources (cpu/mem), is composed of a much complex logic with much more lines of code and I can't see yet the value it's bringing...

felipejfc:
> Looks like your find and replace for the version string included an extra space. I’d approve if I knew more about this code base, but I’m just an onlooker anxious to see this change get merged in case it can help fix the weird Apple TV frame rate judder problems with the Moonlight app.
> 
> Thanks for putting this fix together!

Thanks for reviewing! I've just fixed the typo you pointed out!

travis134:
@felipejfc I setup Xcode and built/deployed to my Apple TV 4K (2nd gen) from your fork. I was able to enable the vsync option but I still get persistent judder which happens every second, and appears to hang both the audio and video streams briefly. I was hoping this vsync option would help but no dice. Is the problem I'm seeing (video attached) what you are seeing on your end or is this a new issue? If so I can create a new issue to document it, or happy to test any future revisions that attempt to address the microstutter issue.

https://user-images.githubusercontent.com/412428/147807217-b7e9a1ea-8f89-4e82-bc05-fb7196f018b7.MOV

felipejfc:
> @felipejfc I setup Xcode and built/deployed to my Apple TV 4K (2nd gen) from your fork. I was able to enable the vsync option but I still get persistent judder which happens every second, and appears to hang both the audio and video streams briefly. I was hoping this vsync option would help but no dice. Is the problem I'm seeing (video attached) what you are seeing on your end or is this a new issue? If so I can create a new issue to document it, or happy to test any future revisions that attempt to address the microstutter issue.
> 
> https://user-images.githubusercontent.com/412428/147807217-b7e9a1ea-8f89-4e82-bc05-fb7196f018b7.MOV

Hey! So I would say you need to check your network setup. I also use an Apple TV 4K 2nd gen and with or without vsync I don't get judder as bad as yours..

Vsync on
https://user-images.githubusercontent.com/285782/147822194-9e07a6e4-9eb5-4d6d-b7cf-81c0034ee7d5.MOV

travis134:
Thanks, yeah I am using Wi-Fi for connectivity. What's strange is if I restart my Apple TV, and start Moonlight immediately after it turns back on I don't see any stutter at all, it's pretty much perfect for a few minutes, after which this constant stutter kicks in and never relents.

I'll try with a wired connection sometime today just to rule out Wi-Fi specifically. If this was network connectivity wouldn't I expect to see frames dropped because of network conditions in the diagnostics overlay? Is there some other way I can debug networking related issues as measured client side?

travis134:
I switched my Wi-Fi channel and used a different network adapter on the host and that seemed to clear up my problem. Sorry for the spam!

cgutman:
@felipejfc Can you try my `pull_renderer_poc` branch? It should accomplish basically the same thing as your branch but it's a bit simpler.

You can experiment with the code in that `#if 0` block. With that code enabled (`#if 1`), it's basically perfectly smooth for me. If that works well for you too, we can wire that up to a setting and ship it as a "frame pacing" option or something.

https://github.com/moonlight-stream/moonlight-ios/blob/512c1479a0adb43da25d102406bea4c79c0d464e/Limelight/Stream/VideoDecoderRenderer.m#L98-L104

felipejfc:
I'm traveling right now so I can't test this, but will do next Tuesday!

But will the result be the same @cgutman? In this branch you're intentionally skipping one frame when you only have 1 enqueued buffer, while in mine I'm always exhausting the queue during each sync step. So maybe mine will carry less delay while possibly being a little less fluid?

Best!

cgutman:
> I'm traveling right now so I can't test this, but will do next Tuesday!

No problem, enjoy your travels.

> In this branch you're intentionally skipping one frame when you only have 1 enqueued buffer, while in mine I'm always exhausting the queue during each sync step. So maybe mine will carry less delay while possibly being a little less fluid?

The 1 frame delay is basically an example of what can be done in that callback. You can write the same pacing logic that your branch has using my branch with less code, or experiment with different techniques to see what works best in your testing.

Rather than having to keep our own queue of buffers to render, as your current branch does, we can use moonlight-common-c's built-in support for pull-based renderers to let it handle the queueing for us. Also since the CADisplayLink callback is happening on the main run loop, we can just render directly from there, instead of needing to dispatch onto the main thread as we do today or needing to wake up a render thread as your branch does.

cgutman:
> The issue is that sometimes there are 2 network frames ready to render for 1 display frame on iPhone 12 Pros. This is because the display framerate on the iPhone 12 is too far from 60 fps. Buffering doesn't solve this issue.

You have that backwards, right? A 60 FPS stream has 1 frame available for every 2 frames needed by the display for 120 Hz.

We shouldn't need to do anything special for 120 Hz, because CADisplayLink will handle it for us. We ask for 60 FPS and it will give us callbacks as if we had a 60 Hz display (or for variable refresh rate displays, just run the display at 60 Hz).

https://github.com/moonlight-stream/moonlight-ios/blob/512c1479a0adb43da25d102406bea4c79c0d464e/Limelight/Stream/VideoDecoderRenderer.m#L82-L84

If we do need to deal with a CADisplayLink running at 120 FPS and a 60 FPS stream, you can easily accomplish this too. Simply render a frame on every other callback (plus whatever other pacing logic you want).

For cases where the user has disabled the frame pacing option for lowest possible latency, we can run the CADisplayLink as fast as possible and render on every callback (saving us 8ms for 60 FPS content on 120 Hz display) at the cost of additional microstutter.

doctorpangloss:
I’m saying that NVIDIA sends frames every 1/60.000 seconds perfectly, with a standard deviation of almost 0.0001s, and the iPhone 12 Pro draws every 1/59.74 to 1/59.76 seconds on average. But an average is an average. Every few seconds there’s as long as 0.030s between CADisplayLink callbacks. Which is still consistent with an average that predicts there should only be frame skipping every 250s. This long tail poor draw rate is the cause of the judder on the iPhone 12 Pro and the Apple TV, and those devices alone, and why there is no judder on other iOS devices, and why you observe no judder at 15fps NVIDIA.

felipejfc:
Even if NVIDIA does try to send at a perfect 60hz you still have statistical fluctuation due to several reasons (processing, networking). If iPhone does draw at 1/59.75 I think it is fine, drawing in the CALink would then be equivalent to doing frame pacing (even without intentionally keeping buffered frames)

doctorpangloss:
> Even if NVIDIA does try to send at a perfect 60hz you still have statistical fluctuation due to several reasons (processing, networking). If iPhone does draw at 1/59.75 I think it is fine, drawing in the CALink would then be equivalent to doing frame pacing (even without intentionally keeping buffered frames) 

I'm saying very often on the iPhone 12 Pro, there is longer than 20ms between draws. So with two frames from the network, one will have to be skipped, and that is observed by judder. You can buffer as many frames as you want but this will not solve the judder due to the iPhone 12 Pro's buggy frame rate. Other users have this issue, including me. You can even observe it in the YouTube app, in the thumbnail previews. The best solution is to blend the frames.

Buffering can only solve network issues. It cannot solve the problem of 2 or more frames wanting to draw at the same time.

furfoots:
CADisplayLink is not handling it properly for the iPhone 12 Pro as all iPhone 12 Pro's refresh at ±59.7x.

felipejfc:
> CADisplayLink is not handling it properly for the iPhone 12 Pro as all iPhone 12 Pro's refresh at ±59.7x. Allowing users to manually input a framerate with preferredFramesPerSecond would fix this.

doesn't this snippet covers for what you're saying? https://github.com/moonlight-stream/moonlight-ios/pull/482/files#diff-d1d948b28ba15e047c42b9575dc4ffcb10fbae5ab4320288f51ff068bb6b81adR90-R92

doctorpangloss:
> doesn't this snippet covers for what you're saying? https://github.com/moonlight-stream/moonlight-ios/pull/482/files#diff-d1d948b28ba15e047c42b9575dc4ffcb10fbae5ab4320288f51ff068bb6b81adR90-R92

To my surprise, asking iOS for an integer 60fps using that API on the iPhone 12 Pro yields an actual refresh rate of 59.74-6 with times between calls as long as 30ms at the 92nd percentile. Meaning, it appears to be an iPhone 12 Pro system-wide bug.

felipejfc:
@cgutman I tested your branch and played for some hours. It's working fine for me! I have reworked the PR renaming "Vsync" to "Frame Pacing" and putting your simpler solution in place, please take a look :)

cgutman:
@felipejfc can you provide some details on what your experience was with the current code, your branch with frame pacing off, and your branch with frame pacing on?

I'm curious if there was any benefit to moving to the CVDisplayLink callback without any other changes.

felipejfc:
> @felipejfc can you provide some details on what your experience was with the current code, your branch with frame pacing off, and your branch with frame pacing on?
> 
> 
> 
> I'm curious if there was any benefit to moving to the CVDisplayLink callback without any other changes.

Honestly, it's very different using it on and off. You can really tell the difference. With it off you feel a lot of tearing, with it on I played the new GOW for like 3 hours and it was very smooth.

I didn't however compare CVDisplayLink with pacing option off vs the current implementation on master to see if the tearing is the same or if it gets worse.

I suggest you try it to see by yourself, for me I was comparing moonlight with pacing on vs steam link, with the pacing on I felt moonlight as smooth or even smoother than SL while still maintaining a little less input delay, I would say they buffer more frames than just one.

cgutman:
OK, I merged my PoC changes to master, so you can rebase your changes on it.

A few CR comments too:
- Please drop the xcscheme files from the PR
- Please use v1.7 for the xcdatamodel. Incrementing it twice to v1.8 is not necessary because v1.7 was never released.
- Please back out the changes the Storyboard files that aren't strictly part of adding the new option

Aljutor:
@felipejfc tested your branch on tvOS (4K second gen), video seems to be more stable, but suddenly there is no audio at all.

felipejfc:
> OK, I merged my PoC changes to master, so you can rebase your changes on it.
> 
> 
> 
> A few CR comments too:
> 
> - Please drop the xcscheme files from the PR
> 
> - Please use v1.7 for the xcdatamodel. Incrementing it twice to v1.8 is not necessary because v1.7 was never released.
> 
> - Please back out the changes the Storyboard files that aren't strictly part of adding the new option

Will do! I'm traveling again but will be able to do it during the next weekend.

felipejfc:
> @felipejfc tested your branch on tvOS (4K second gen), video seems to be more stable, but suddenly there is no audio at all.
> 
> No sure is this a problem with this branch exactly, or with mainstream changes in audio. Going to check It later. 

Weird, I did rebase the branch with the SDL audio changes @cgutman pushed in the last week. But I tried it and played for like 3 hours straight in my 4K 2nd gen Apple TV with no problems at all.

Aljutor:
> > @felipejfc tested your branch on tvOS (4K second gen), video seems to be more stable, but suddenly there is no audio at all.
> > No sure is this a problem with this branch exactly, or with mainstream changes in audio. Going to check It later.
> 
> Weird, I did rebase the branch with the SDL audio changes @cgutman pushed in the last week. But I tried it and played for like 3 hours straight in my 4K 2nd gen Apple TV with no problems at all.

Checked mainstream master, no audio to.  Maybe I building project somehow wrongly.

Aljutor:
Ok, I temporary created my own fork with old audio, but cherry-picked new pull-based renderer and frame pacing on top of it.  Seems to work fine for now. But big disclaimer, I am not so much experienced with both Xcode/tvOS and overall graphic programming.  It's only for testing now, and cherrypicking overrides @felipejfc as author of some commits (I am not claiming any rights obviously).  

If someone can review this changes, I can provides updates to @felipejfc PR to make it suitable for merging.   
You can find it at - https://github.com/Aljutor/moonlight-ios/tree/old_audio

cgutman:
It is not necessary to revert the SDL audio changes for this PR, it's a known issue that will be addressed separately.

Aljutor:
> It is not necessary to revert the SDL audio changes for this PR, it's a known issue that will be addressed separately.

I done this only to test for audio/video desync. In this PR indeed it is not necessary.

Aljutor:
@felipejfc @cgutman I rebased this branch onto current master, tested it for some time (apple tv4k, 4k stream, 150 bitrate). For me it seems that frame pacing helps a lot with sudden small stutters. I can create separate PR for my fork, or @felipejfc can pull changes (if he wish of course) to his. Not sure what easier to do, and unfortunately Github does not provide a nice way to suggest changes to someone PR if you are not a repo maintainer.

neatnoise:
I'm having micro stuttering issue as well on ios (iPhone 12 mini) whatever settings i set (HEVC ON/OFF, different resolutions, bitrate). I use sunshine for the host.
Steamlink on ios is smooth without issues.

I have no issues with micro stuttering on MacOS with the recent 3.2.0 release (vsync and frame pacing ON), it's smooth there

Aljutor:
@felipejfc In case you too busy now.  If you ok with this - I can open new PR with your changes and fixes for PR ontop of it to speedup things.

felipejfc:
> @felipejfc In case you too busy now. If you ok with this - I can open new PR with your changes and fixes for PR ontop of it to speedup things.

don't worry, I'm busy with work but I will be able to do the small changes needed this week

felipejfc:
@cgutman just rebased and made the changes you asked for! please take a look

PR #486, opened 2022-01-24
Use actual number of audio channels

### LINKED ISSUE #484: No audio after switching to SDL on AppleTV 4K (tvOS 15.2)
Tested current master [bd582aa](https://github.com/moonlight-stream/moonlight-ios/commit/bd582aa6c0c0e4db283c166395ef3d362201fcc6) on Apple TV 4K (2 gen) with tvOS 15.2 and found out that there no audio. Version from AppStore works fine. To additionally clarify issue I branched before SDL2 commits [8391c76](https://github.com/moonlight-stream/moonlight-ios/commit/8391c766c7545f6d041528dfbbaa4a9c9ecd34b6) cherry picked bd582aa onto it and tested this version, audio works fine.

Aljutor:
Fixes #484 On Apple TV maximumOutputNumberOfChannels reports 5.1 audio even when actually there is only stereo audio enabled. This leads to having input from PC being processed as 5.1 even when device is not capable of doing so and ultimately to having no audio.

### DIFF
--- a/Limelight/ViewControllers/MainFrameViewController.m
+++ b/Limelight/ViewControllers/MainFrameViewController.m
@@ -631,8 +631,19 @@ static NSMutableSet* hostList;
     _streamConfig.multiController = streamSettings.multiController;
     _streamConfig.gamepadMask = [ControllerSupport getConnectedGamepadMask:_streamConfig];
     
+    // Probe for supported channel configurations.
+    // First try to switch audio route to max output channels is possible
+    AVAudioSession *audio_instance = [AVAudioSession sharedInstance];
+    long maxOutputChannels = audio_instance.maximumOutputNumberOfChannels;
+    Log(LOG_I, @"Audio device supports %d channels", maxOutputChannels);
+    
+    bool sucsess = [audio_instance setPreferredOutputNumberOfChannels:maxOutputChannels error:nil];
+    if (sucsess && [audio_instance outputNumberOfChannels] == maxOutputChannels){
+        Log(LOG_I, @"Switched audio device to %d channels", maxOutputChannels);
+    } else {
+        Log(LOG_I, @"Failed to switch audio device to %d channels", maxOutputChannels);
+    }
 
-    // Probe for supported channel configurations
     long outputChannels = [AVAudioSession sharedInstance].outputNumberOfChannels;
     Log(LOG_I, @"Audio device supports %d channels", outputChannels);
     if (outputChannels >= 8) {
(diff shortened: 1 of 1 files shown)


cgutman:
I'm not totally convinced this is correct. [Kodi's audio sink](https://github.com/xbmc/xbmc/blame/ebf1c8ac591a0dd09ac6f55fcef3a45193140a87/xbmc/cores/AudioEngine/Sinks/AESinkDARWINTVOS.mm) seems to do the same thing we are doing here.

My understanding of the relationship between `maximumOutputNumberOfChannels` and `outputNumberOfChannels` is that `maximumOutputNumberOfChannels` reports the maximum number of channels supported by the current audio route and `outputNumberOfChannels` reports the actual negotiated number of channels in use (which can be influenced by `setPreferredOutputNumberOfChannels`).

It doesn't seem to explain what was reported [here](https://github.com/moonlight-stream/moonlight-ios/issues/371#issuecomment-1013618952) where the user clearly has a 5.1 surround sound system (and can use it successfully with Steam Link). Moonlight also chose 5.1, but output was still broken.

Aljutor:
@cgutman I think we have two separate issues, one is not working 5.1 and other - no stereo audio. 
Pushed new changes, now client will try to switch for max audio channels first, and will use actual number after.  
Basically same as VLC https://github.com/videolan/vlc/blob/def4dea4dc7470aa6cfc96a4ca99975f041d4954/modules/audio_output/audiounit_ios.m#L229

felipejfc:
When I tried this here, In mediaserverd I get logs saying it will use 8 channels, maximumOutputNumberOfChannels  reports 32 then it tries to set preferredOutputNumberOfChannels to it, which fails (I also tried to force setting 8 and it also fails) and then when I print outputNumberOfChannels I get only 2
I have ATV 4k 2nd gen connected directly to a dolby atmos capable soundbar

Aljutor:
> When I tried this here, In mediaserverd I get logs saying it will use 8 channels, maximumOutputNumberOfChannels reports 32 then it tries to set preferredOutputNumberOfChannels to it, which fails (I also tried to force setting 8 and it also fails) and then when I print outputNumberOfChannels I get only 2
> I have ATV 4k 2nd gen connected directly to a dolby atmos capable soundbar

Can you check that in other software on appletv that it works in 8 channels? Maybe we should use not the maximum but what mediaserverd supports.

felipejfc:
I can get Dolby Atmos in Netflix, Disney, etc...

cgutman:
I decided to just provide an option in settings for now, so the user can pick whatever works for them.

PR #503, opened 2022-04-10
feat: If the frame will be immediately override, do not show it

felipejfc:
Part of what was done in #501. In the same DisplayLink tick its no use to display more than 1  frame, so we ignore all but the last


felipejfc:
Closing this PR. I will open another one to discuss another idea

Issue #511, opened 2022-07-03
Enable client-side button remapping on iOS Moonlight client


IonBlade2K:
**Is your feature request related to a problem? Please describe.**
iOS/iPadOS 16 support Nintendo Switch Joy-Cons.  However, the iOS button mapping is setup in a way that swaps A and B, X and Y:

Joy-Con Input -> iOS button Output
A (right face button) -> A (bottom face button)
B (bottom face button) -> B (right face button)
X (top face button) -> X (left face button)
Y (left face button) -> Y (top face button)

Unfortunately, the native Joy-Con combining does not support iOS's native controller profile behavior for button remapping, so the buttons cannot be remapped at the input level in the OS.  While they can be remapped in Steam on the remote side, some games, like Destiny 2, do not support Steam's button remapping as an anti-cheat measure, so global remapping at the PC side also isn't an option. 

It seems as if Apple intends for this to be the behavior of Joy-Cons in iOS/iPadOS 16 - they want the button with "A" written on it to send "A" as output, even if that's not the same physical location button on Xbox controllers, and intend for developers of apps to enable remapping to the proper output button if that functionality is needed, meaning this would have to be fixed either client-side in Moonlight or remapped PC-side (which not all games support).

**Describe the solution you'd like**
A clean global fix to this would be to enable button remapping within the Moonlight iOS client - while a full client-side remapping option would be great, an easy fix for the specific Joy-Con issue would be a toggle in Moonlight to simply swap A<->B and X<->Y.  This would enable Switch Joy-Cons to have the proper output to PC in iOS 16, so pressing the "A" button on the Joy-Con (right face button) would send "B" to the virtual Xbox controller (right face button), to not have to press different buttons than one would on a real controller, or rely on PC-side remaps per game, which not all games support.

**Describe alternatives you've considered**

- I've tried using the two Joy-Cons as separate controllers through the iOS16 native combination of Joy-Cons (Joy-Cons can be combined into one controller or separated into two separate controllers in iOS16 natively by holding Home + Capture for a few seconds), then using Buddy Controller to combine them, instead of the native iOS combination.  This does not work, because Buddy Controller does not allow for iOS gamepad remapping profiles to be used to enable the remapping.

[...]

PR #513, opened 2022-07-08
Add setting to swap A<->B and X<->Y buttons

IonBlade2K:
iOS 16 adds support for Nintendo Joy-Cons, including a combined mode where both Joy-Cons appear as one controller.  Apple has chosen to implement this in a way where pressing a button on the Joy-Con sends the button with the same _letter_, but not the same _location_ on an Xbox controller:

Joy-Con Input -> iOS button Output / Moonlight Xbox Controller Output
A (right face button) -> A (bottom face button)
B (bottom face button) -> B (right face button)
X (top face button) -> X (left face button)
Y (left face button) -> Y (top face button)

This leads to having to re-learn muscle memory or remapping inputs PC-side per game (which not all games support) when using Joy-Cons / Switch Pro Controllers.

Apple seems to want applications to do the remapping where outputs should be swapped to match physical Xbox button _locations_.; combined Joy-Cons at an iOS system level do not support the built-in iOS button remapping features in Settings -> Game Controllers to enable this mapping to be done at a system level.

This commit adds a new toggle to Moonlight iOS to add an A/B and X/Y swap toggle in Settings, so that users of Joy-Cons can optionally swap A<->B and X<->Y in Moonlight settings.  This enables using Joy-Cons / Pro Controller with the same muscle memory as Xbox / PS controllers, without any PC side remapping needed.

Implements the minimum viable product needed to fix the enhancement for Joy-Con support requested in https://github.com/moonlight-stream/moonlight-ios/issues/511.

### DIFF
--- a/Limelight/Database/DataManager.h
+++ b/Limelight/Database/DataManager.h
@@ -21,6 +21,7 @@
                 onscreenControls:(NSInteger)onscreenControls
                    optimizeGames:(BOOL)optimizeGames
                  multiController:(BOOL)multiController
+                 swapABXYButtons:(BOOL)swapABXYButtons
                        audioOnPC:(BOOL)audioOnPC
                          useHevc:(BOOL)useHevc
                   useFramePacing:(BOOL)useFramePacing
--- a/Limelight/Database/DataManager.m
+++ b/Limelight/Database/DataManager.m
@@ -60,6 +60,7 @@
                 onscreenControls:(NSInteger)onscreenControls
                    optimizeGames:(BOOL)optimizeGames
                  multiController:(BOOL)multiController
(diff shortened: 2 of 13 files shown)


cgutman:
Thanks, there are a few minor issues, but I will fix them after merging.

IonBlade2K:
Thanks!  Sorry for the handful of issues; I’ve never touched iOS dev or Xcode before. Just worked backwards from “I know what the gui prompts are, let’s find where the existing toggle text is referenced and trace backwards through the code to try to add a similar feature.” :-)

cgutman:
No worries. That was pretty well done for your first time doing iOS dev.

PR #521, opened 2022-08-16
Added 1440p resolution option

Starlank:
Was originally done by [pierguil](https://github.com/pierguil), but it looks like their branch is outdated.
The branch in question: https://github.com/pierguil/moonlight-ios/tree/1440p

Never contributed to an iOS project before, would love any feedback.

### DIFF
--- a/Limelight/Database/TemporarySettings.m
+++ b/Limelight/Database/TemporarySettings.m
@@ -60,6 +60,10 @@
             self.height = [NSNumber numberWithInteger:2160];
             self.width = [NSNumber numberWithInteger:3840];
             break;
+        case 3:
+            self.height = [NSNumber numberWithInteger:1440];
+            self.width = [NSNumber numberWithInteger:2560];
+            break;
         default:
             abort();
     }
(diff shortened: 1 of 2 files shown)


cgutman:
Let's limit this to tvOS (which lacks custom resolution support) only

iOS can use the custom resolution option to use 1440p already. Due to GeForce Experience behaving badly with resolutions that aren't 720p, 1080p, or 4K, I'd like users to go through the UX flow that displays the custom resolution warning dialog on iOS.

Limelight/Database/TemporarySettings.m
```diff
@@ -57,6 +57,10 @@ - (id) initFromSettings:(Settings*)settings {
             self.width = [NSNumber numberWithInteger:1920];
             break;
         case 2:
+            self.height = [NSNumber numberWithInteger:1440];
+            self.width = [NSNumber numberWithInteger:2560];
+            break;
+        case 3:
```
cgutman, line 63: It's not safe to use 2 for 1440p, because 2 is what is currently used for 4K. You should use 3 for 1440p instead.

cgutman (changes requested):

Starlank:
Does this look better? Isolated it to just tvOS, and changed 1440p to option 3.

cgutman (approved):
Perfect, thanks.

Issue #491, opened 2022-02-08
Mouse speed way too slow on iPad


rayvearn:
I use the magic keyboard, and the mouse speed is insanely slow, even after changing the mouse sensitivity to 100 on windows and the iPad. Please add an option to increase it when streaming or at least an automation to max it out while you are streaming. No idea why it’s this slow, might just have to do with some iOS/iPadOS weirdness.

ericsyj:
I encountered the same situation when using the Bluetooth mouse on my iPad Pro 2021. Are there any improvements in subsequent versions?

ooijiyoung:
Same issue using Bluetooth Anywhere MX2 with iPad Air Gen 4

basedcumrade:
Same problem every time with the Magic Keyboard trackpad

beren12:
Same with the Logitech Combo trackpad. I don’t recall this issue using a Bluetooth mouse a year or 2 ago

ashathe:
Is this issue being improved? I’m still encountering the issue.

oxkenshin:
Same issue on my iPad Magic Keyboard trackpad.

oxkenshin:
I used to fixed this by change MOUSE_SPEED_DIVISOR from 2.5 to 1.0 in "Limelight/Input/ControllerSupport.m". But I am not sure if this will cause issues for other input devices other than magic keyboard trackpad. It would be good to add an option to change this.

clock145:
mark

cgutman:
I don't have the hardware to test this right now. Can someone else try the changes in #532 to confirm they address the issue?

RacerDelux:
> I don't have the hardware to test this right now. Can someone else try the changes in #532 to confirm they address the issue?

If I had a mac to build the app on =(... Were those changes committed and published already?

rayvearn:
i don't have a mac to build the app with the change :/

lightsabata:
> I don't have the hardware to test this right now. Can someone else try the changes in #532 to confirm they address the issue?

After building and testing i can confirm that it resolve the issue pretty well !

Client : iPadOS 16.2
Build with Xcode 14.2

PS: i will be a nice enhancement if it can be set manually on the client settings (horizontal scrollbar on left panel ?) since Windows and macOS hosts have not the same mouse sensitivity (via sunshine)

PR #532, opened 2022-11-03
Fix mouse speed too slow on iPad

oxkenshin:


oxkenshin:
#491

oxkenshin:
MOUSE_SPEED_DIVISOR = 1.25 is speedy enough

PR #535, opened 2022-11-26
[WIP] Use vt for manually decoding frames. Fixes #533

### LINKED ISSUE #533: Apple TV (2022 Version) HDR Black Screen
The 2022 version of the Apple TV displays a black screen when doing an HDR stream. My 2021 Apple TV has no issues on the same TV, connected to the same computer. SDR streams continue to work fine on the 2022 Apple TV.

felipejfc:
Two main changes

1 - Use VideoToolbox to manually decode each frame instead of submitting it directly to AVSampleBufferDisplayLayer; I'm not proud of this change, but it was needed to fix https://github.com/moonlight-stream/moonlight-ios/issues/533. There may be some way to fix the issue without needing this change, but I still didn't manage to do it.

2 - Latency and smoothness changes
2.1 - Use Direct Submit in VideoDecodeRenderer (reduces latency)
2.2 - Use PTS information correctly per frame instead of using the DisplayImmediately flag in each sampleBuffer. Together with the change above, I was able to replicate smooth low latency stream as I get into the Nvidia Shield. I think using the flag messed with frame time and caused jittering.

Right now, I'm breaking the "Smooth Stream" option that we added some months ago, but wanted to create the PR either way for us to discuss options @cgutman

### DIFF
--- a/Limelight/Stream/VideoDecoderRenderer.m
+++ b/Limelight/Stream/VideoDecoderRenderer.m
@@ -6,6 +6,8 @@
 //  Copyright (c) 2014 Moonlight Stream. All rights reserved.
 //
 
+@import VideoToolbox;
+
 #import "VideoDecoderRenderer.h"
 #import "StreamView.h"
 
@@ -23,6 +25,7 @@
     
     NSData *spsData, *ppsData, *vpsData;
     CMVideoFormatDescriptionRef formatDesc;
+    VTDecompressionSessionRef decompressionSession;
     
     CADisplayLink* _displayLink;
     BOOL framePacing;
@@ -74,6 +77,12 @@
         CFRelease(formatDesc);
         formatDesc = nil;
     }
+    
+    if (decompressionSession != nil){
+        VTDecompressionSessionInvalidate(decompressionSession);
+        CFRelease(decompressionSession);
+        decompressionSession = nil;
+    }
 }
 
 - (id)initWithView:(StreamView*)view callbacks:(id<ConnectionCallbacks>)callbacks streamAspectRatio:(float)aspectRatio useFramePacing:(BOOL)useFramePacing
@@ -94,10 +103,7 @@
 {
     self->videoFormat = videoFormat;
     self->frameRate = frameRate;
-}
-
-- (void)start
-{
+    
     _displayLink = [CADisplayLink displayLinkWithTarget:self selector:@selector(displayLinkCallback:)];
     if (@available(iOS 15.0, tvOS 15.0, *)) {
         _displayLink.preferredFrameRateRange = CAFrameRateRangeMake(self->frameRate, self->frameRate, self->frameRate);
@@ -106,6 +112,26 @@
         _displayLink.preferredFramesPerSecond = self->frameRate;
     }
     [_displayLink addToRunLoop:[NSRunLoop mainRunLoop] forMode:NSDefaultRunLoopMode];
+
+}
+
+- (void) setupDecompressionSession {
+    if (decompressionSession != NULL){
+        VTDecompressionSessionInvalidate(decompressionSession);
+        CFRelease(decompressionSession);
+        decompressionSession = nil;
+    }
+    
+    int status = VTDecompressionSessionCreate(kCFAllocatorDefault,
+                                             formatDesc,
+                                             nil,
+                                             nil,
+                                             nil,
+                                             &decompressionSession);
+    if (status != noErr) {
+        NSLog(@"Failed to instance VTDecompressionSessionRef, status %d", status);
+    }
+
 }
 
 // TODO: Refactor this
@@ -113,31 +139,10 @@ int DrSubmitDecodeUnit(PDECODE_UNIT decodeUnit);
 
 - (void)displayLinkCallback:(CADisplayLink *)sender
 {
-    VIDEO_FRAME_HANDLE handle;
-    PDECODE_UNIT du;
-    
-    while (LiPollNextVideoFrame(&handle, &du)) {
-        LiCompleteVideoFrame(handle, DrSubmitDecodeUnit(du));
-        
-        if (framePacing) {
-            // Calculate the actual display refresh rate
-            double displayRefreshRate = 1 / (_displayLink.targetTimestamp - _displayLink.timestamp);
-            
-            // Only pace frames if the display refresh rate is >= 90% of our stream frame rate.
-            // Battery saver, accessibility settings, or device thermals can cause the actual
-            // refresh rate of the display to drop below the physical maximum.
-            if (displayRefreshRate >= frameRate * 0.9f) {
-                // Keep one pending frame to smooth out gaps due to
-                // network jitter at the cost of 1 frame of latency
-                if (LiGetPendingVideoFrames() == 1) {
-                    break;
-                }
-            }
-        }
-    }
+    // Do Nothing
 }
 
-- (void)stop
+- (void)cleanup
 {
     [_displayLink invalidate];
 }
@@ -262,6 +267,8 @@ int DrSubmitDecodeUnit(PDECODE_UNIT decodeUnit);
                     formatDesc = NULL;
                 }
             }
+            
+            [self setupDecompressionSession];
         }
         
         // Data is NOT to be freed here. It's a direct usage of the caller's buffer.
@@ -330,11 +337,12 @@ int DrSubmitDecodeUnit(PDECODE_UNIT decodeUnit);
         
     CMSampleBufferRef sampleBuffer;
     
-    status = CMSampleBufferCreate(kCFAllocatorDefault,
+    CMSampleTimingInfo sampleTiming = {kCMTimeInvalid, CMTimeMake(pts, 1000), kCMTimeInvalid};
+    
+    status = CMSampleBufferCreateReady(kCFAllocatorDefault,
                                   frameBlockBuffer,
-                                  true, NULL,
-                                  NULL, formatDesc, 1, 0,
-                                  NULL, 0, NULL,
+                                  formatDesc, 1, 1,
+                                  &sampleTiming, 0, NULL,
                                   &sampleBuffer);
     if (status != noErr) {
         Log(LOG_E, @"CMSampleBufferCreate failed: %d", (int)status);
@@ -342,40 +350,88 @@ int DrSubmitDecodeUnit(PDECODE_UNIT decodeUnit);
         CFRelease(frameBlockBuffer);
         return DR_NEED_IDR;
     }
+       
+    OSStatus decodeStatus = [self decodeFrameWithSampleBuffer: sampleBuffer frameType: frameType];
     
-    CFArrayRef attachments = CMSampleBufferGetSampleAttachmentsArray(sampleBuffer, YES);
-    CFMutableDictionaryRef dict = (CFMutableDictionaryRef)CFArrayGetValueAtIndex(attachments, 0);
-    
-    CFDictionarySetValue(dict, kCMSampleAttachmentKey_DisplayImmediately, kCFBooleanTrue);
-    CFDictionarySetValue(dict, kCMSampleAttachmentKey_IsDependedOnByOthers, kCFBooleanTrue);
-    
-    if (frameType == FRAME_TYPE_PFRAME) {
-        // P-frame
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanTrue);
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanTrue);
+    if (decodeStatus != noErr){
+        NSLog(@"Failed to decompress frame");
     } else {
-        // I-frame
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanFalse);
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanFalse);
+ 
     }
+    
+    /* Flush in-process frames. */
+    //VTDecompressionSessionFinishDelayedFrames(decompressionSession);
+    
+    /* Block until our callback has been called with the last frame. */
+    //VTDecompressionSessionWaitForAsynchronousFrames(decompressionSession);
 
-    // Enqueue the next frame
-    [self->displayLayer enqueueSampleBuffer:sampleBuffer];
-    
-    if (frameType == FRAME_TYPE_IDR) {
-        // Ensure the layer is visible now
-        self->displayLayer.hidden = NO;
-        
-        // Tell our parent VC to hide the progress indicator
-        [self->_callbacks videoContentShown];
-    }
-    
     // Dereference the buffers
     CFRelease(dataBlockBuffer);
     CFRelease(frameBlockBuffer);
     CFRelease(sampleBuffer);
     
     return DR_OK;
+}
+
+- (OSStatus) decodeFrameWithSampleBuffer:(CMSampleBufferRef)sampleBuffer frameType:(int)frameType{
+    VTDecodeFrameFlags flags = kVTDecodeFrame_EnableAsynchronousDecompression;
+    VTDecodeInfoFlags flagOut = 0;
+    
+    return VTDecompressionSessionDecodeFrameWithOutputHandler(decompressionSession, sampleBuffer, flags, &flagOut, ^(OSStatus status, VTDecodeInfoFlags infoFlags, CVImageBufferRef  _Nullable imageBuffer, CMTime presentationTimestamp, CMTime presentationDuration) {
+        if (status != noErr)
+        {
+            NSError *error = [NSError errorWithDomain:NSOSStatusErrorDomain code:status userInfo:nil];
+            NSLog(@"Decompression session error: %@", error);
+        }
+        
+        CMVideoFormatDescriptionRef formatDescriptionRef;
+        
+        OSStatus res = CMVideoFormatDescriptionCreateForImageBuffer(kCFAllocatorDefault, imageBuffer, &formatDescriptionRef);
+        if (res != noErr){
+            NSLog(@"Failed to create video format description from imageBuffer");
+        }
+        
+        CMSampleBufferRef sampleBuffer;
+        CMSampleTimingInfo sampleTiming = {kCMTimeInvalid, presentationTimestamp, kCMTimeInvalid};
+        
+        OSStatus err = CMSampleBufferCreateReadyWithImageBuffer(kCFAllocatorDefault, imageBuffer, formatDescriptionRef, &sampleTiming, &sampleBuffer);
+        
+        if (err != noErr){
+            NSLog(@"Error creating sample buffer for decompressed image buffer %d", (int)err);
+            return;
+        }
+        
+        CFArrayRef attachments = CMSampleBufferGetSampleAttachmentsArray(sampleBuffer, YES);
+        CFMutableDictionaryRef dict = (CFMutableDictionaryRef)CFArrayGetValueAtIndex(attachments, 0);
+        
+        CFDictionarySetValue(dict, kCMSampleAttachmentKey_IsDependedOnByOthers, kCFBooleanTrue);
+        
+        if (frameType == FRAME_TYPE_PFRAME) {
+            // P-frame
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanTrue);
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanTrue);
+        } else {
+            // I-frame
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanFalse);
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanFalse);
+        }
+        
+        // Enqueue the next frame
+        [self->displayLayer enqueueSampleBuffer:sampleBuffer];
+        
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (frameType == FRAME_TYPE_IDR) {
+                // Ensure the layer is visible now
+                self->displayLayer.hidden = NO;
+                
+                // Tell our parent VC to hide the progress indicator
+                [self->_callbacks videoContentShown];
+            }
+        });
+        
+        CFRelease(sampleBuffer);
+        CFRelease(formatDescriptionRef);
+    });
 }
 
 @end
--- a/Limelight/Stream/VideoDecoderRenderer.m
+++ b/Limelight/Stream/VideoDecoderRenderer.m
@@ -79,6 +79,11 @@
         formatDesc = nil;
     }
     
+    if (formatDescImageBuffer != nil) {
+        CFRelease(formatDescImageBuffer);
+        formatDescImageBuffer = nil;
+    }
+    
     if (decompressionSession != nil){
         VTDecompressionSessionInvalidate(decompressionSession);
         CFRelease(decompressionSession);
@@ -350,21 +355,6 @@ int DrSubmitDecodeUnit(PDECODE_UNIT decodeUnit);
         CFRelease(dataBlockBuffer);
         CFRelease(frameBlockBuffer);
         return DR_NEED_IDR;
-    }
-    
-    CFArrayRef attachments = CMSampleBufferGetSampleAttachmentsArray(sampleBuffer, YES);
-    CFMutableDictionaryRef dict = (CFMutableDictionaryRef)CFArrayGetValueAtIndex(attachments, 0);
-    
-    CFDictionarySetValue(dict, kCMSampleAttachmentKey_IsDependedOnByOthers, kCFBooleanTrue);
-    
-    if (frameType == FRAME_TYPE_PFRAME) {
-        // P-frame
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanTrue);
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanTrue);
-    } else {
-        // I-frame
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanFalse);
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanFalse);
     }
        
     OSStatus decodeStatus = [self decodeFrameWithSampleBuffer: sampleBuffer frameType: frameType];
--- a/Limelight/Stream/VideoDecoderRenderer.m
+++ b/Limelight/Stream/VideoDecoderRenderer.m
@@ -25,6 +25,7 @@
     
     NSData *spsData, *ppsData, *vpsData;
     CMVideoFormatDescriptionRef formatDesc;
+    CMVideoFormatDescriptionRef formatDescImageBuffer;
     VTDecompressionSessionRef decompressionSession;
     
     CADisplayLink* _displayLink;
@@ -129,7 +130,7 @@
                                              nil,
                                              &decompressionSession);
     if (status != noErr) {
-        NSLog(@"Failed to instance VTDecompressionSessionRef, status %d", status);
+        Log(LOG_E, @"Failed to instance VTDecompressionSessionRef, status %d", status);
     }
 
 }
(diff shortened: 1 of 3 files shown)


Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
     return DR_OK;
 }
 
+- (OSStatus) decodeFrameWithSampleBuffer:(CMSampleBufferRef)sampleBuffer frameType:(int)frameType{
+    VTDecodeFrameFlags flags = kVTDecodeFrame_EnableAsynchronousDecompression;
```
cgutman, line 377: Does async decompression result in improved performance vs synchronous?

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+    VTDecodeInfoFlags flagOut = 0;
+    
+    return VTDecompressionSessionDecodeFrameWithOutputHandler(decompressionSession, sampleBuffer, flags, &flagOut, ^(OSStatus status, VTDecodeInfoFlags infoFlags, CVImageBufferRef  _Nullable imageBuffer, CMTime presentationTimestamp, CMTime presentationDuration) {
```
cgutman, line 380: `VTDecodeInfoFlags` are optional, so we can just pass `NULL`, since we never read them.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+        if (status != noErr)
+        {
+            NSError *error = [NSError errorWithDomain:NSOSStatusErrorDomain code:status userInfo:nil];
+            NSLog(@"Decompression session error: %@", error);
```
cgutman, line 384: We should call `LiRequestIdrFrame()` and return here.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+        }
+        
+        CMVideoFormatDescriptionRef formatDescriptionRef;
+        
+        OSStatus res = CMVideoFormatDescriptionCreateForImageBuffer(kCFAllocatorDefault, imageBuffer, &formatDescriptionRef);
+        if (res != noErr){
+            NSLog(@"Failed to create video format description from imageBuffer");
```
cgutman, line 391: Please change the `NSLog()` calls to our `Log()` function instead.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -330,46 +337,34 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
-        self->displayLayer.hidden = NO;
-        
-        // Tell our parent VC to hide the progress indicator
-        [self->_callbacks videoContentShown];
-    }
+    /* Flush in-process frames. */
+    //VTDecompressionSessionFinishDelayedFrames(decompressionSession);
     
+    /* Block until our callback has been called with the last frame. */
+    //VTDecompressionSessionWaitForAsynchronousFrames(decompressionSession);
```
cgutman, line 366: If these aren't needed, please remove them.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -330,46 +337,34 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
-    
-    CFDictionarySetValue(dict, kCMSampleAttachmentKey_DisplayImmediately, kCFBooleanTrue);
-    CFDictionarySetValue(dict, kCMSampleAttachmentKey_IsDependedOnByOthers, kCFBooleanTrue);
-    
-    if (frameType == FRAME_TYPE_PFRAME) {
-        // P-frame
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanTrue);
-        CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanTrue);
+    if (decodeStatus != noErr){
+        NSLog(@"Failed to decompress frame");
```
cgutman, line 357: We should return `DR_NEED_IDR` here.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+        CFDictionarySetValue(dict, kCMSampleAttachmentKey_IsDependedOnByOthers, kCFBooleanTrue);
+        
+        if (frameType == FRAME_TYPE_PFRAME) {
+            // P-frame
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanTrue);
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanTrue);
+        } else {
+            // I-frame
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_NotSync, kCFBooleanFalse);
+            CFDictionarySetValue(dict, kCMSampleAttachmentKey_DependsOnOthers, kCFBooleanFalse);
```
cgutman, line 416: These attributes should be set on the H.264/HEVC `CMSampleBuffer` that we pass to the `VTDecompressionSession` rather than the `CMSampleBuffer` we pass to `AVBufferSampleDisplayLayer` (which is now just raw YUV data).

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+        
+        CMVideoFormatDescriptionRef formatDescriptionRef;
+        
+        OSStatus res = CMVideoFormatDescriptionCreateForImageBuffer(kCFAllocatorDefault, imageBuffer, &formatDescriptionRef);
+        if (res != noErr){
+            NSLog(@"Failed to create video format description from imageBuffer");
+        }
+        
+        CMSampleBufferRef sampleBuffer;
+        CMSampleTimingInfo sampleTiming = {kCMTimeInvalid, presentationTimestamp, kCMTimeInvalid};
```
cgutman, line 395: Even though we don't set `duration` ourselves in our VT sample timing info, we should probably pass `presentationDuration` in here from the VT callback just in case.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +373,65 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+    return VTDecompressionSessionDecodeFrameWithOutputHandler(decompressionSession, sampleBuffer, flags, &flagOut, ^(OSStatus status, VTDecodeInfoFlags infoFlags, CVImageBufferRef  _Nullable imageBuffer, CMTime presentationTimestamp, CMTime presentationDuration) {
+        if (status != noErr)
+        {
+            NSError *error = [NSError errorWithDomain:NSOSStatusErrorDomain code:status userInfo:nil];
+            NSLog(@"Decompression session error: %@", error);
+        }
+        
+        CMVideoFormatDescriptionRef formatDescriptionRef;
+        
+        OSStatus res = CMVideoFormatDescriptionCreateForImageBuffer(kCFAllocatorDefault, imageBuffer, &formatDescriptionRef);
```
cgutman, line 389: Can we cache this `CMVideoFormatDescriptionRef` to avoid having to recreate it for each frame? `CMVideoFormatDescriptionMatchesImageBuffer()` should tell us if we need to create a new one again.

cgutman (changes requested):
These changes are looking pretty good.

I think this has the potential to improve the frame pacing option too. Now that we have access to the decoded samples, we can keep a queue of those and submit those in our display link callback. Since we were queuing compressed video samples before, there was the potential that we could miss a frame deadline if the decoder couldn't finish decoding that frame by the time it was due to be displayed.

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: Honestly, I didn't compare sync/async here because I couldn't find a reliable way to measure performance other than my gut feeling. Any ideas?

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: makes sense!

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: I was in doubt regarding leaving or removing; the reason I left is it sounds like this option can be useful in the future, but I can remove it

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: makes total sense! will change it

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: think it always receives .timeInvalid, but can change it

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: sure! reason I left it recreating is that I saw some Apple's samples doing the same and I believe it's a "cheap" op; but can change it

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: should we also call LiRequestIdrFrame() here?

felipejfc:
Plot Twist: As part of tackling the improvements you suggested @cgutman, I hit the reason for the original issue #533. Still unsure about the root cause, but it's these lines:

https://github.com/moonlight-stream/moonlight-ios/blob/master/Limelight/Stream/VideoDecoderRenderer.m#L346-L360

I ran some tests, and we don't even need to set any value in the dict; only getting it will cause the decoder to go nuts:

```
   CFArrayRef attachments = CMSampleBufferGetSampleAttachmentsArray(sampleBuffer, YES);
    CFMutableDictionaryRef dict = (CFMutableDictionaryRef)CFArrayGetValueAtIndex(attachments, 0);
```

This is enough to make the decoder fail when the sample buffers contain HDR data. I imagine it must be due to some OS bug.

Given that:
- Do you think the solution to decouple decoding still makes sense? Or should we move forward with removing these lines only? Honestly, I'm not sure if they make any difference. From my tests, I see none when I change them.
- I can send another PR with only this change and using the PTS information; Since we can't get the reference to the dict this will be needed as we can't set DisplayImmediately flag. WDYT?

cgutman:
> Do you think the solution to decouple decoding still makes sense? Or should we move forward with removing these lines only? Honestly, I'm not sure if they make any difference. From my tests, I see none when I change them.

Removing those lines should be fine. 

> I can send another PR with only this change and using the PTS information; Since we can't get the reference to the dict this will be needed as we can't set DisplayImmediately flag. WDYT?

Yep, let's do that to fix #533 ASAP and we can see if using a VTDecompressionSession improves things further vs the current pure AVSampleBufferDisplayLayer solution.

Do you see a frame pacing regression using the PTS info with the pacing option enabled? If so, we can just use this solution for HDR streaming only for now.

felipejfc:
@cgutman as part of the changes, I wanted to test different ways of decoding and rendering, using VT to manually decode and update a CALayer with the resulting image, continue to pass the encoded buffer directly to AVSampleBufferDisplayLayer, and then measure the latency of each approach.
Any ideas on how I could benchmark these solutions reliably?

cgutman:
I suppose you could use a phone in slow motion mode.

For now though, let's try to get HDR on the new Apple TV, then we can fine tune things later.

Starlank:
@felipejfc do the latency improvements only apply to the 2022 Apple TV 4K?

felipejfc:
@Starlank the changes here and in the other PR should not improve latency; but should improve stream "smoothness" using the low latency pacing mode. I'm currently studying latency improvements locally.

Limelight/Stream/VideoDecoderRenderer.m
Alanko5: Not every frame is decompressed asynchronously, this is the correct setting to increase speed.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -106,38 +118,37 @@ - (void)start
         _displayLink.preferredFramesPerSecond = self->frameRate;
     }
     [_displayLink addToRunLoop:[NSRunLoop mainRunLoop] forMode:NSDefaultRunLoopMode];
+
+}
+
+- (void) setupDecompressionSession {
+    if (decompressionSession != NULL){
+        VTDecompressionSessionInvalidate(decompressionSession);
```
Alanko5, line 126: it is also necessary to do the following:
VTDecompressionSessionWaitForAsynchronousFrames(decompressionSession);

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +372,52 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+        
+        OSStatus err = CMSampleBufferCreateReadyWithImageBuffer(kCFAllocatorDefault, imageBuffer, self->formatDescImageBuffer, &sampleTiming, &sampleBuffer);
+        
+        if (err != noErr){
+            Log(LOG_E, @"Error creating sample buffer for decompressed image buffer %d", (int)err);
+            return;
+        }
+        
+        // Enqueue the next frame
+        [self->displayLayer enqueueSampleBuffer:sampleBuffer];
```
Alanko5, line 407: I would add a flush before the [self->displayLayer enqueueSampleBuffer:sampleBuffer] function.
``` 
if (![self->displayLayer isReadyForMoreMediaData]) {
        [self->displayLayer flush];
}
```
Sometimes it happens that not all data is played, and the buffer fills up, so the playback stops.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -378,4 +372,52 @@ - (int)submitDecodeBuffer:(unsigned char *)data length:(int)length bufferType:(i
...
+            
+            OSStatus res = CMVideoFormatDescriptionCreateForImageBuffer(kCFAllocatorDefault, imageBuffer, &(self->formatDescImageBuffer));
+            if (res != noErr){
+                Log(LOG_E, @"Failed to create video format description from imageBuffer");
+                return;
+            }
+        }
+        
+        CMSampleBufferRef sampleBuffer;
+        CMSampleTimingInfo sampleTiming = {kCMTimeInvalid, presentationTimestamp, presentationDuration};
```
Alanko5, line 397: I recommend using CACurrentMediaTime() for timing info.
In this way, the frames will be played immediately.

Limelight/Stream/VideoDecoderRenderer.m
```diff
@@ -106,38 +118,37 @@ - (void)start
...
+}
+
+- (void) setupDecompressionSession {
+    if (decompressionSession != NULL){
+        VTDecompressionSessionInvalidate(decompressionSession);
+        CFRelease(decompressionSession);
+        decompressionSession = nil;
+    }
+    
+    int status = VTDecompressionSessionCreate(kCFAllocatorDefault,
```
Alanko5, line 131: For initialize VTDecompressionSession need set parameters.
If I understand it correctly, the sender can send data from various sources.
Therefore, it is necessary to prepare for the fact that the data can be of various types.
For example:
```
        let imageBufferAttributes = [
            //kCVPixelBufferPixelFormatTypeKey: NSNumber(value: kCVPixelFormatType_......), /// if needed
            kCVPixelBufferIOSurfacePropertiesKey: [:] as AnyObject, 
            kCVPixelBufferOpenGLESCompatibilityKey: NSNumber(booleanLiteral: true), 
            kCVPixelBufferMetalCompatibilityKey: NSNumber(booleanLiteral: true), 
            kCVPixelBufferOpenGLCompatibilityKey: NSNumber(booleanLiteral: true) 
        ]
```

Alanko5:
If you like SDR change to HDR:
```
let pixelTransferProperties = [kVTPixelTransferPropertyKey_DestinationColorPrimaries: kCVImageBufferColorPrimaries_ITU_R_2020,
                                           kVTPixelTransferPropertyKey_DestinationTransferFunction: kCVImageBufferTransferFunction_SMPTE_ST_2084_PQ,
                                           kVTPixelTransferPropertyKey_DestinationYCbCrMatrix: kCVImageBufferYCbCrMatrix_ITU_R_2020]

VTSessionSetProperty(decompressionSession,
                                 key: kVTDecompressionPropertyKey_PixelTransferProperties,
                                 value: pixelTransferProperties as CFDictionary)
```
Do not forget that on tvOS it is necessary to switch the TV to HDR mode.

Limelight/Stream/VideoDecoderRenderer.m
felipejfc: I have the impression if I set it to display immediately, more jittering is generated, maybe because presentationDuration gets messed up?

Alanko5: set the duration if you know the fps.

Alanko5: but try it without setting the duration, I don't notice the jittering in another project.

felipejfc:
Thanks for the review @Alanko5. I have doubts regarding the manual decompression approach though, as I wasn't able to reduce video latency.
The way that reduced it the most was using kCVPixelBufferIOSurfacePropertiesKey property so that the image received in the decompression callback has a backing IOSurface and setting the displayLayer contents directly (ditching the SampleBufferDisplayLayer basically)
Btw given how much latency the AppleTV, even newest model has, when compared to ipads or iphones I think that it's some hardware related latency between the ATV and the display (monitor/TV)

felipejfc:
> If you like SDR change to HDR:
> 
> ```
> let pixelTransferProperties = [kVTPixelTransferPropertyKey_DestinationColorPrimaries: kCVImageBufferColorPrimaries_ITU_R_2020,
>                                            kVTPixelTransferPropertyKey_DestinationTransferFunction: kCVImageBufferTransferFunction_SMPTE_ST_2084_PQ,
>                                            kVTPixelTransferPropertyKey_DestinationYCbCrMatrix: kCVImageBufferYCbCrMatrix_ITU_R_2020]
> 
> VTSessionSetProperty(decompressionSession,
>                                  key: kVTDecompressionPropertyKey_PixelTransferProperties,
>                                  value: pixelTransferProperties as CFDictionary)
> ```
> 
> Do not forget that on tvOS it is necessary to switch the TV to HDR mode.

Is this SDR->HDR mapping?

Alanko5:
> > If you like SDR change to HDR:
> > ```
> > let pixelTransferProperties = [kVTPixelTransferPropertyKey_DestinationColorPrimaries: kCVImageBufferColorPrimaries_ITU_R_2020,
> >                                            kVTPixelTransferPropertyKey_DestinationTransferFunction: kCVImageBufferTransferFunction_SMPTE_ST_2084_PQ,
> >                                            kVTPixelTransferPropertyKey_DestinationYCbCrMatrix: kCVImageBufferYCbCrMatrix_ITU_R_2020]
> > 
> > VTSessionSetProperty(decompressionSession,
> >                                  key: kVTDecompressionPropertyKey_PixelTransferProperties,
> >                                  value: pixelTransferProperties as CFDictionary)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do not forget that on tvOS it is necessary to switch the TV to HDR mode.
> 
> Is this SDR->HDR mapping?

Yes, apple mentions it somewhere in the documentation. 
It does not generate an HDR image, but it improves SDR.

Alanko5:
> Thanks for the review @Alanko5. I have doubts regarding the manual decompression approach though, as I wasn't able to reduce video latency.
> The way that reduced it the most was using kCVPixelBufferIOSurfacePropertiesKey property so that the image received in the decompression callback has a backing IOSurface and setting the displayLayer contents directly (ditching the SampleBufferDisplayLayer basically)
> Btw given how much latency the AppleTV, even newest model has, when compared to ipads or iphones I think that it's some hardware related latency between the ATV and the display (monitor/TV)

I don't think it's caused by HW.
tvOS is a different system than iOS.
I think that it is enough to find some setting that will only be enabled.
But I may be wrong.

Alanko5:
> Thanks for the review @Alanko5. I have doubts regarding the manual decompression approach though, as I wasn't able to reduce video latency.
> The way that reduced it the most was using kCVPixelBufferIOSurfacePropertiesKey property so that the image received in the decompression callback has a backing IOSurface and setting the displayLayer contents directly (ditching the SampleBufferDisplayLayer basically)
> Btw given how much latency the AppleTV, even newest model has, when compared to ipads or iphones I think that it's some hardware related latency between the ATV and the display (monitor/TV)

What latency are we talking about?
Now I tried to measure the decoding time.
H265 decoding took 0.001sec.
Where do you see this delay?
Maybe I don't fully understand the problem.

felipejfc:
> > Thanks for the review @Alanko5. I have doubts regarding the manual decompression approach though, as I wasn't able to reduce video latency.
> > The way that reduced it the most was using kCVPixelBufferIOSurfacePropertiesKey property so that the image received in the decompression callback has a backing IOSurface and setting the displayLayer contents directly (ditching the SampleBufferDisplayLayer basically)
> > Btw given how much latency the AppleTV, even newest model has, when compared to ipads or iphones I think that it's some hardware related latency between the ATV and the display (monitor/TV)
> 
> What latency are we talking about? Now I tried to measure the decoding time. H265 decoding took 0.001sec. Where do you see this delay? Maybe I don't fully understand the problem.

There's streaming delay witn ATV4K when compared to streaming with an iphone/ipad or nvidia shield. I compared them using a stopwatch application and slow-mo iPhone camera to compare PC screen time with streaming screen time

Alanko5:
Oh, I understand.
Measure how long it takes you to decode.
According to my measurements, it is 0.001sec.
If you think the delay is causing the decoding, switch to H264.
There, the delay is even smaller.

In my opinion, the timing will help you solve the problem.

felipejfc:
> I understand. Measure how long it takes you to decode. According to my measurements, it is 0.001sec. If you think the delay is causing the decoding, switch to H264. There, the delay is even smaller.
> 
> In my opinion, the timing will help you solve the problem.
> 
> Do you not use WiFi when measuring? :-)

For 4k HEVC I think I was getting 8ms time to decode each frame

Alanko5:
Did you measure the decompression time of the Key and non-Key frames?
Can you set the server to send fewer keyframes? (for example one per two seconds)

What is the total delay of the image that you measured with the camera?

What version of apple tv do you have?

How do you create a VTDecompressionSession?
I mean, what parameters are you setting?

felipejfc:
Code is in this branch https://github.com/felipejfc/moonlight-ios/tree/ds_queue_surface

I have the latest 4K Apple TV(2022) with the iPhone 12 pro processor.

_What is the total delay of the image that you measured with the camera?_ 
The streamed image would be always 25~50 hundredths behind the original image; when testing with m1 MacBook, or nvidia shield, most of the time the images would be in sync.

_Did you measure the decompression time of the Key and non-Key frames?_
I measured all frames and they all took this same amnt of time.

_Can you set the server to send fewer keyframes?_

Pretty sure gamestream won't allow me to do it

Alanko5:
According to what you write, the problem is not in decoding.
I think that by improving the decoding you can gain a maximum of 5ms.
The first thing I would look for is a network or rendering delay.
Because a delay of 250~500ms is huge!

Ak zistis ze vykreslovanie je pomale.
Este je tu moznost pouzit vlastny MetalView.

Well, you can try as follows:

It is necessary that you set this value (As I wrote above):
kCVPixelBufferMetalCompatibilityKey

Your destinationImageBufferAttributes:
```
NSDictionary *pixelAttributes = @{
        (id)kCVPixelBufferMetalCompatibilityKey : (id)kCFBooleanTrue,
        (id)kCVPixelBufferIOSurfaceCoreAnimationCompatibilityKey : (id)kCFBooleanTrue,
        (id)kCVPixelBufferIOSurfacePropertiesKey : @{},
    };
```
I think that during rendering it would help if the layer could use Metal.


```
NSDictionary *videoDecoderSpec = @{
         (id) kCMFormatDescriptionExtension_FullRangeVideo : FORMAT_DESC_FullRangeVideo,
         (id) kCVImageBufferChromaLocationBottomFieldKey: kCVImageBufferChromaLocation_Left,
         (id) kCVImageBufferChromaLocationTopFieldKey: kCVImageBufferChromaLocation_Left,
         (id) kCVImageBufferPixelAspectRatioKey: FORMAT_DESC_AspectRatio,
         (id) kCVImageBufferColorPrimariesKey: FORMAT_DESC_ColorPrimaries,
         (id) kCVImageBufferTransferFunctionKey: FORMAT_DESC_TransferFunction,
         (id) kCVImageBufferYCbCrMatrixKey: FORMAT_DESC_YCbCrMatrix,
         if (HEVC) {
               (id) kVTVideoDecoderSpecification_EnableHardwareAcceleratedVideoDecoder : (id)kCFBooleanTrue
         } else {
               (id) kVTVideoDecoderSpecification_RequireHardwareAcceleratedVideoDecoder : (id)kCFBooleanTrue
         }
};
```

felipejfc:
> According to what you write, the problem is not in decoding. I think that by improving the decoding you can gain a maximum of 5ms. The first thing I would look for is a network or rendering delay. Because a delay of 250~500ms is huge!
> 
> Well, you can try as follows:
> 
> It is necessary that you set this value (As I wrote above): kCVPixelBufferMetalCompatibilityKey
> 
> Your destinationImageBufferAttributes:
> 
> ```
> NSDictionary *pixelAttributes = @{
>         (id)kCVPixelBufferMetalCompatibilityKey : (id)kCFBooleanTrue,
>         (id)kCVPixelBufferIOSurfaceCoreAnimationCompatibilityKey : (id)kCFBooleanTrue,
>         (id)kCVPixelBufferIOSurfacePropertiesKey : @{},
>     };
> ```
> 
> I think that during rendering it would help if the layer could use Metal.
> 
> ```
> NSDictionary *videoDecoderSpec = @{
>          (id) kCMFormatDescriptionExtension_FullRangeVideo : FORMAT_DESC_FullRangeVideo,
>          (id) kCVImageBufferChromaLocationBottomFieldKey: kCVImageBufferChromaLocation_Left,
>          (id) kCVImageBufferChromaLocationTopFieldKey: kCVImageBufferChromaLocation_Left,
>          (id) kCVImageBufferPixelAspectRatioKey: FORMAT_DESC_AspectRatio,
>          (id) kCVImageBufferColorPrimariesKey: FORMAT_DESC_ColorPrimaries,
>          (id) kCVImageBufferTransferFunctionKey: FORMAT_DESC_TransferFunction,
>          (id) kCVImageBufferYCbCrMatrixKey: FORMAT_DESC_YCbCrMatrix
> };
> ```

Sorry, I misspelled it. It's actually 25~50ms delay! I will try your changes anyways when I get home; travelling right now so that's only next week

jasin755:
Why is this still not resolved? What is it waiting for?

andygrundman:
A few of us have been hacking on some iOS frame pacing stuff and I've incorporated this PR into some test code I'm working on. My branch is at https://github.com/andygrundman/moonlight-ios/tree/andyg.ios-frame-pacing but fair warning it's very much an experimental WIP. Just wanted you to know this good patch has not been forgotten.

pierreski:
I'm available to help with testing to close out this issue as well! Just got an Apple TV 4K, and the amount of lag I'm experiencing in Moonlight is very noticeable.

- I experience no perceptible lag when using an Nvidia Shield, but the lag on Apple TV 4K is noticeable
- I've tried using both an Xbox one controller connected to both the Apple TV 4K and the host computer - noticeable lag in both scenarios
- My Apple TV 4K is connected via ethernet, so bandwidth / network latency is not the problem

adamsondavid:
Hi, same here. I also just got a new Apple TV 4k 3rd gen experiencing render delays. The input device is directly connected to the PC. I would also be happy to offer my help on testing just like @pierreski

PhilSpieltLOL:
The render delays are still an issue for me as well. I ended up having to switch to a FireTV Cube just for Streaming because of that. My input device is connected to my PC directly as well through a VirtualHere Setup. I would be happy to help with testing as well.

andygrundman:
FYI a few months ago I implemented a Metal renderer along with direct VT decoding for moonlight-ios, and this feature has since made its way into the VoidLink iOS app. VT decoding is always active, the Metal renderer is optional and marked as experimental. It's unfortunately not a free app, but just letting you know the feature is implemented if you want it.

PR #539, opened 2022-12-06
Adding more Bitrate options to tvOS

HakanFly:
Added bitrate value options of 70 and 80 Mbps in the Root.plist file specific to Moonlight TV.


Starlank:
Definitely in support of this!

PR #552, opened 2023-02-21
Update bug and feature_request forms

bladeoner:
@cgutman I have updated the reports the same as I did for Android but these are tailored for iOS.


bladeoner:
@cgutman when you have some time can you take a look at the PR?

Thanks in advance.

Issue #547, opened 2023-01-16
[IpadOS] Plz support external display full screen mode


BLOODTYP:
Please use the airplay method like the full screen of the ott such as Netflix, or add a function that supports the full screen of the external display stage manager.

cd419:
Yes I just tried this and hit the limitation of not being able to scale the app full screen when using stage manager and an external display. I’d like to take advantage of this to do some gaming on an external display which could be really cool.

leobeosab:
It's been a while since I've done IOS dev but I'm running into this issue as well. I'll see if I can get stage manager + full screen working on an external display.

leobeosab:
I was able to get fullscreen working on an external display by remove the requires full screen attribute from Limlight-Info.plist. It works fine for me with no drops in performance but I don't know if this would negatively affect other users.

![Screenshot 2023-02-23 at 2 37 56 PM](https://user-images.githubusercontent.com/1686221/221014473-9fde2281-964a-44d5-8750-827b9e8ed0e6.png)
Example of it working connected to a dock:
![03CA8D5D-1DE2-4F6C-A59F-E6E6BFE2CEBE](https://user-images.githubusercontent.com/1686221/221015207-bc7cc9d8-1ac7-408a-a98e-3f8c4ae61e65.jpg)

BLOODTYP:
> It's been a while since I've done IOS dev but I'm running into this issue as well. I'll see if I can get stage manager + full screen working on an external display.

Thank you, but I don't know how to develop. How can I release this code as an official app?

cd419:
@leobeosab Thanks for taking a look at this! Would love to be able to help test this but haven't delved into dev on iOS but I don't have a mac. Is there a way to build this directly on an iPad or windows / linux PC?

patroza:
this may be the only thing keeping me from sunshine/moonlight, and leaves me stuck on steamlink

ThuRoldao:
Any news on this matter? I would also love to be able to use the app in fullscreen on my external monitor

jessedrakehovey:
Any news on this? Haven't seen a comment since 2024

PR #553, opened 2023-02-23
Issue #547 fix fullscreen on iPad external monitor, remove UIRequires…

leobeosab:
I removed the UIRequiresFullScreen to be able to go fullscreen and resize the app on an external display, I tested it on my iPad and my phone (not long on the phone though to be honest) and it seemed to work fine. 

I am a backend / devops person so let me know if this will break current functionality or anything.

03CA8D5D-1DE2-4F6C-A59F-E6E6BFE2CEBE

### DIFF
--- a/Limelight/Limelight-Info.plist
+++ b/Limelight/Limelight-Info.plist
@@ -73,7 +73,7 @@
 		<string>armv7</string>
 	</array>
 	<key>UIRequiresFullScreen</key>
-	<true/>
+	<false/>
 	<key>UIStatusBarHidden</key>
 	<true/>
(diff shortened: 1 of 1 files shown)


empireslayer000:
Hello, please follow up with this and pull this into the master, seems to work according to the above and I have tested this myself.

WolfLink:
This wasn't enough from my testing, but adding "Portrait" and "Upside Down" orientations as well as removing the "Requires Fullscreen" did.

Also, it looks like there is an API for external monitors that might be interesting for implementing a more thorough solution: https://developer.apple.com/documentation/uikit/windows_and_screens/presenting_content_on_a_connected_display.

PR #557, opened 2023-03-19
iPadOS External Display Fullscreen Support

empireslayer000:
Simple fork of most recent release with the change made to support fullscreen external displays.

This was brought up in a previous pull request but the user closed it out.
Hope to have something like this rolled out in the next IOS release.

Cheers.

### DIFF
--- a/Limelight/Limelight-Info.plist
+++ b/Limelight/Limelight-Info.plist
@@ -69,7 +69,7 @@
 		<string>armv7</string>
 	</array>
 	<key>UIRequiresFullScreen</key>
-	<true/>
+	<false/>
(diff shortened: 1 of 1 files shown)


empireslayer000:
@cgutman any follow up to this?

sapoepsilon:
I wanted to implement it myself, but looks like @empireslayer000 has already implemented it. @empireslayer000  Maybe calling reviewers will speed up the merge process? @cgutman

patroza:
how is this helping? can this only be used with stagemanager and when using a mouse?
steam link keeps the main app on iPad screen and projects full screen game on external screen - this would imo be the best and most amazing solution

PR #576, opened 2023-08-29
Add a toolbar above the on-screen keyboard with extra keys that are missing from iOS keyboard

goob47:
This is a fairly simple addition to the keyInputField that shows the keyboard when a three finger tap is detected. Now a toolbar with more buttons such as Windows, Control, Alt, Delete, Shift, Escape, and Tab is available so if you need to quit out of a crashed game or menu, you can use Alt-Tab in the toolbar to get back to your desktop. 

There is a toggle for these keys: Windows, Shift, Control, Alt -- this allows them to be used as modifier keys and you can trigger most keyboard shortcuts now such as Windows-Tab, Control-W, etc. The only issue that I encountered was when trying to use Control-Alt-Delete. I think this is a limitation with the LiSendKeyboardEvent function not being able to send three key events concurrently. I tried a few methods to manually trigger it if the combo is detected but couldn't figure it out. 

Please let me know if you have any questions or concerns with my code. I am new to Obj-C so I apologize if my conventions are all over the place. 

I love this software, thank you so much for all your work on this and making it free and open-source!


goob47:
Here is how the feature looks in action:
 
screenshot

cgutman:
Merged, thanks.

Btw, Ctrl+Alt+Del doesn't work not because of any issue with Moonlight, but because it's the secure attention sequence. It cannot be injected via virtual input APIs like `SendInput()`.

PR #577, opened 2023-08-30
Add support for full screen and safe area resolutions

goob47:
Another new feature I've been wanting to add to the app -- pretty self-explanatory. I have also added a view below the resolution switcher that will show the full width x height in pixels that the user will need to change their PC's display to. 

public

When the resolution display view is tapped, the user is redirected to this Nvidia support page which shows how to create a custom resolution: [https://nvidia.custhelp.com/app/answers/detail/a_id/759/~/custom-resolutions](https://nvidia.custhelp.com/app/answers/detail/a_id/759/~/custom-resolutions)

This feature allows users to utilize their full display's resolution if their PC and game supports custom resolutions. This was already possible with the Custom option, but now you don't need to Google your device's resolution every time. Now gameplay like this is made easy on notched devices:

IMG_1102

Please let me know if you have any questions about the feature. It probably needs adjustments for non-notched devices and Apple TV.


Heeervas:
This is just incredible. THANKS!

PR #600, opened 2023-12-15
Add fn 1-12  to toolbar above keyboard to support screen switch

oxkenshin:
Screenshot 2023-12-15 at 4 55 41 PM

Here is how it looks alike.

This modification made it possible to switch screen from iOS with software keyboard using the shortcut ctrl-shift-alt-fn[1-6].


Limelight/Input/StreamView.m
```diff
@@ -356,7 +356,7 @@ - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
                 keyInputField.text = @"0";
 #if !TARGET_OS_TV
                 // Prepare the toolbar above the keyboard for more options
-                UIToolbar *customToolbarView = [[UIToolbar alloc] initWithFrame:CGRectMake(0, 0, self.bounds.size.width, 44)];
+                UIToolbar *customToolbarView = [[UIToolbar alloc] initWithFrame:CGRectMake(0, 0, self.bounds.size.width > 1500 ? self.bounds.size.width : 1500, 44)];
```
cgutman, line 359: How did you come up with 1500 here?

oxkenshin: By test :(
Didn't find a solution to calculate the width before drawing. I am not familiar with programing on iOS. Any suggestion?

oxkenshin: > How did you come up with 1500 here?

New pull request has removed this magic number by correctly calculating the toolbar width. For completeness, it also adds F7-F12 to the toolbar.
The new UI layout also makes the "cross" button on another level to make it always visible.

oxkenshin:
Updated Toolbar UI. Tested on iPad Pro 12.9 inch, and iPad mini 6.

Screenshot 2023-12-30 at 9 33 56 PM
IMG_0072

oxkenshin:
Adds arrow key to support shortcuts like Win+Shift+Left/Right Arrow to switch a window to left/right screen.
Screenshot 2023-12-31 at 2 27 00 PM

knaos:
This is great and a very needed functionality. I definately need it a lot. Alternatively, can there also be a special button like Next display just for switching the displays? Or a gesture, like three fingers swipe or 4 fingers swipe?

medicalwei:
Am looking for this to be merged, as iPad keyboard case does not have Fn keys and I need it for my use case.

What criteria are needed for verification?

Issue #515, opened 2022-07-12
iPadOS 16 stage manager and external display support 


cooldna2:
Now that iPadOS 16 public beta’s been released, I tried opening moonlight on external display being controlled by stage manager (thunderbolt 3), and the app window is rotated by 90 degrees. I don’t know but it seems to be an issue that can be fixed simply I hope..

minor-coder:
What happens if I turn on the moonlight on the iPad screen after connecting an external monitor?

cooldna2:
On the iPad screen the app runs without any issue…

kurozero0:
Looks like the rotation was fixed in recent iOS betas. I would like to see full screen support on external monitors, like the PlayStation remote play app.

Intertwined4:
Yes, please add fullscreen support.

caccialdo:
Seconding the feature request for full screen support 🙇 🙏

chrispouliot:
I’m enjoying using moonlight on my iPad, but the fact that it stays at the iPad’s 4:3 (?) aspect ratio when using Stage Manager is annoying. Allowing a 16:9 ratio to go full screen on my external monitor would be really nice. Thanks!

KleistyLoczek:
Please add this! Its so annoying to see 3:2 aspect ratio instead of 16:9...

cheveeth:
Please!! It would be a dream come true if we can enter full screen mode on iPad stage manager on external display!!

Issue #444, opened 2020-12-31
After ipad pro connecting the monitor,  The screen is not full screen


gjc9620:

AlexKuai:
Full screen

kxjko:
Some app can play video on monitor with full screen, like tv app. I think this can be used for reference.

ReclusiveRychuRychu:
I'm interested to see if this changes with version 8.0, if not there's gotta be a way to initiate airplay just like Youtube or Netflix does.

kxjko:
@ReclusiveRychuRychu Thank you for focussing on this problem. It’s the same in 8.0.

PR #411, opened 2020-05-20
Adding External Screen Support [Feature Request #11]

mattrussell7:
Added External Screen Support for all iOS devices whether through HDMI Out via Lightning/USB-C or Airplay Screen Mirroring. 

Feature Request #11 discusses this.

PR #608, opened 2024-02-12
Enable external display support on iPad by allowing the OS to control the window shape and size.

WolfLink:
Removing "UIRequiresFullScreen" and adding all 4 of the orientations will cause the OS to enable all of the iPadOS multitasking configurations, which most importantly includes going full screen on an external display.

This will also allow having another app open side-by-side with this one on the iPad itself.


WolfLink:
Here are issues with people requesting this behavior:

- https://github.com/moonlight-stream/moonlight-ios/issues/547
- https://github.com/moonlight-stream/moonlight-ios/issues/515
- https://github.com/moonlight-stream/moonlight-ios/issues/444

MrSunshineVS:
This is the last piece of the puzzle for me. Would love to see this in the next update!

cgutman:
This can't be merged without properly handling rotation and resizing while streaming. The stream view, on screen controls, and stats overlay all get cut off if resized or rotation while streaming.

WolfLink:
Just wanted to add that I have started working on this.  I have a proof-of-concept, but I don't want to update the pull request until I think its ready.

TheTawnyFool:
This is what I need also to use Moonlight. My process won't allow me to not use it on an external display. Excited to see this feature, and thank you so much for addressing it!

cd419:
Thanks for trying to implement this! If there is anything I can do regarding testing etc. I would love to help out any way I can.

N4ir0lf:
I´m happy that i found that here, thought already i´m using it wrong :D Thanks for that work. This would make the setup just perfect

Vidukka:
Any news on this? Would love to use my iPad on TV with moonlight

twilightgod:
Please make it work! Thank you

lprhodes:
Probably need to lock the orientation while steaming to avoid a lot of issues with rotation.

WolfLink:
So I’ve got a version that supports rotating mid-stream no problem.  The delay is that the on-screen controls don’t scale properly to arbitrary resolutions so I need to fix that.  Also there’s something weird with the computer selection GUI that I haven’t looked into yet.

Rayray1099:
Any new updates on this? Thanks :)

228peace:
@WolfLink  thank u for doing this, love to see this in the next update!

twilightgod:
Hope to see this feature coming soon, thank you

Hung1610 (approved):

BaoLunia:
I signed up just to say If this gets merged i will be eternally grateful.

andygrundman:
I was surprised this was only a few lines but after trying it out, it doesn't appear to work. Can anyone tell me what I'm doing wrong?

iPad Pro M4 -> Thunderbolt Dock -> USB-C to Displayport to a 4K monitor. After pairing a keyboard and mouse it will let me use it as a proper external display, but the Moonlight window is only ever able to take up an iPad-aspect square-ish area in the center of the screen. Clicking the 3 dots -> Enter Full Screen does nothing, and you can't drag the window because the aspect ratio can't be changed. I tried both with and without Stage Manager. To be fair I never use my iPad this way so I might be missing a setting or something. I tried to make a screen cap video but it wouldn't record the external display unfortunately.

Does anyone in this thread have comments about the other external display patch from way back in 2020? https://github.com/moonlight-stream/moonlight-ios/pull/411

WolfLink:
Sorry its been so long.  I primarily made this change for my own usage and didn't have much time to do the extra work to try to get this merged.

I've now made it so it refreshes relevant UI sizes whenever the window changes size.  Controls, stream view, and stream selector will all resize properly.

It's not perfect, and at the smallest window sizes, the on-screen-controls will get a little cut off and at the smallest size opening the settings will cut off the view such that you can't close it without making the view bigger.  But fixing these would require significant reworking of the on-screen-controls and settings layouts.


But I'm hoping these are small enough edge cases to be ignored for now.  Let me know!

WolfLink:
> I was surprised this was only a few lines but after trying it out, it doesn't appear to work. Can anyone tell me what I'm doing wrong?
> 
> iPad Pro M4 -> Thunderbolt Dock -> USB-C to Displayport to a 4K monitor. After pairing a keyboard and mouse it will let me use it as a proper external display, but the Moonlight window is only ever able to take up an iPad-aspect square-ish area in the center of the screen. Clicking the 3 dots -> Enter Full Screen does nothing, and you can't drag the window because the aspect ratio can't be changed. I tried both with and without Stage Manager. To be fair I never use my iPad this way so I might be missing a setting or something. I tried to make a screen cap video but it wouldn't record the external display unfortunately.
> 
> Does anyone in this thread have comments about the other external display patch from way back in 2020? #411

This sounds to me like you built the app targeting an older version of iPadOS than the one your iPad is running.  Apple does this sometimes, I'm guessing as a way to ensure an app "kinda works" even if it wasn't updated to include support for whatever the latest UI changes are, even though from my experience it usually makes the app worse.

WolfLink:
> This can't be merged without properly handling rotation and resizing while streaming. The stream view, on screen controls, and stats overlay all get cut off if resized or rotation while streaming.

@cgutman Please take another look and let me know if you'd like me to keep working on on-screen-controls and settings UI or if this is good enough.

andygrundman:
@WolfLink I hacked a bit on this and I guess it's quite a different approach. I had to refactor a lot to use the iOS 13 Scene API. This branch needs more work but it does let you use USB-C displays on iPhone/iPad as well as AirPlay the stream (obviously too laggy for most things but it was a neat feature you get for free). 

My branch: https://github.com/andygrundman/moonlight-ios/tree/andyg.ipad-external-display

Vladoker:
Will these changes be made with the Iphone or only with the Ipad?

WolfLink:
This commit brings the changes to iPad only (I don’t think iPhone supports the same external monitor behavior), but it should be easy to bring the changes to iPhone.

Vladoker:
> This commit brings the changes to iPad only (I don’t think iPhone supports the same external monitor behavior), but it should be easy to bring the changes to iPhone.

I'm not good at mobile development, but it's pretty easy to assemble a file and upload it to an iphone. Can I ask you to please make changes for the iphone so that it is possible to open moonlight in full-screen format as shadow does?

https://www.youtube.com/shorts/hMsyUypLtF8

WolfLink:
Interesting. I don’t think my changes will enable this on iPhone. I am relying on iPadOS’s “Stage Manager”. These changes simply let Stage Manager resize the window to whatever it wants. The iPhone is different. What my changes would do on iPhone is let you run Moonlight in portrait mode.

You might have better luck with @andygrundman’s fork https://github.com/andygrundman/moonlight-ios/tree/andyg.ipad-external-display


I don’t have a USBC iPhone so I didn’t even think to try this.

Issue #428, opened 2020-11-01
Globe key as escape key


cooldna2:
When moonlight is streaming a game, let there be an option that makes globe key on an ipad keyboard function as escape key. That would be very handy. There is global option which makes globe key as escape key, but I want this is an app-specific function, because when texting I want globe key function as emoji which is iOS 14 default as far as I know.

don17sch:
@cooldna2 You can already set the globe key to be escape from iPadOS settings. When it is set to escape in iPadOS Moonlight will treat it as a escape key

WolfLink:
I made a pull request that does this (https://github.com/moonlight-stream/moonlight-ios/pull/610).  Yes you can change it in iPad OS settings, but since moonlight can't use the Globe key anyway, it seems reasonable to map that key to something more useful.  That way if a user doesn't want to remap that key globally, they can still have it be useful in moonlight.

My pull request maps it to Escape but it might be useful to have it be customizable (e.g. see https://github.com/moonlight-stream/moonlight-ios/issues/356)

PR #610, opened 2024-02-12
Map the "Globe" / "Language" key to Escape

WolfLink:
The "Globe" / "Language" key, which is the bottom-left most key on Apple branded iPad keyboards, is currently unused (it falls through the switch statement).

This commit maps that key to "Escape", which is an important key for many games and other applications, but is not present on most Apple branded iPad keyboards.


WolfLink:
This feature has been requested in this issue: https://github.com/moonlight-stream/moonlight-ios/issues/428

Issue #597, opened 2023-12-14
Multip


Qian23333:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

TrueZhuangJia:
Just came from bilibili, request the same feauture because of the newly found GenshinImpact PC touch UI

leverule:
Request support for multi touch. Remote streaming Genshin by touch mode, really great, original experience.
Even with HDR support

HibanaMA:
Really need multi-touch to use my iPad as a sub-monitor. And it would bring better gaming experience if can edit the virtual game pad individually(like adding extra keyboard button then adjust the size and position freely).

xyaoo:
This is really important, I hope to add it

CalvinNg-git:
I can't agree more

TimmyOVO:
The Moonlight client already incorporates numerous multi-finger gestures, such as triple-tap to open the keyboard and pinch-to-zoom for resizing the StreamView. Adding multi-touch passthrough functionality would likely cause severe conflicts with existing gestures. Personally, I believe it's more appropriate to refine  these existing multi-touch controls before we implementing the multi-touch passthrough.

TrueZhuangJia:
Hi guys, I created a fork to implement multi-touch on iOS
https://github.com/TrueZhuangJia/moonlight-ios-NativeMultiTouchPassthrough/tree/Moonlight-iOS-ZWM

TimmyOVO:
> The Moonlight client already incorporates numerous multi-finger gestures, such as triple-tap to open the keyboard and pinch-to-zoom for resizing the StreamView. Adding multi-touch passthrough functionality would likely cause severe conflicts with existing gestures. Personally, I believe it's more appropriate to refine these existing multi-touch controls before we implementing the multi-touch passthrough.

I already working on a fork added the multiple improvements and new functionality including multi-touch passthrough, u guys might want [check it out ](https://github.com/TimmyOVO/moonlight-ios-NE)

ChrisKimZHT:
> I already working on a fork added the multiple improvements and new functionality including multi-touch passthrough, u guys might want [check it out ](https://github.com/TimmyOVO/moonlight-ios-NE)

I have no macOS devices and couldn't complie this code. Could you please add a .ipa releases? That's will help me a lot!

TrueZhuangJia:
[...]

Issue #596, opened 2023-12-12
Touch Control Improvement


StaticDoDoubleG:
**Is your feature request related to a problem? Please describe.**
Moonlight on iOS lacks proper touch control.
I cannot scroll by touch in iPad. These touch function works properly at Android Environment, so I think this is an iOS version app problem.

**Describe the solution you'd like**
Please make iOS app compatible with windows touch control.
Erase the useless zoom function, I think this function might bother proper touch control.
If you cannot erase it, please make an option to choose it. 

**Additional context**
This video is a comparison between iOS and Android Moonlight App. 
You can see iOS version doesn't support proper touch control, while Android version fully supports touch function of windows.

https://github.com/moonlight-stream/moonlight-ios/assets/133185833/c42a39bf-987b-4040-a732-b0caf2f1ed59

uEternali:
SAME ISSUE

TrueZhuangJia:
Hi guys, I created a fork to implement multi-touch on iOS. Have a try if you are able to sideload the app on iOS
https://github.com/TrueZhuangJia/moonlight-ios-NativeMultiTouchPassthrough/tree/Moonlight-iOS-ZWM

Issue #518, opened 2022-07-28
Is it possible to scroll when touch mode is set to touchscreen?


swhoro:
Now when I use touch screen mode and use two fingers, it just moves the picture or zoom in and zoom out it.

laralove143:
Also experiencing this

uEternali:
same issue

Issue #496, opened 2022-03-08
Picture in picture mode, touch mode, mouse delay


BAByte:
1. Picture in picture mode: once the application switches to the background and enters again, it needs to be reconnected, which is very unreasonable. Can you add picture in picture mode?

2. Touch mode: if the iPad can support touch, it is excellent.

3. UX: sliding along the left edge will immediately disconnect. I don't think this operation is a good interactive logic. I suggest using buttons or shortcut keys.

4. Mouse delay: this should be a pain point. I hope it can be solved.

cgutman:
> Touch mode: if the iPad can support touch, it is excellent.

Did you adjust the settings to use touchscreen mode instead of trackpad mode?

> Mouse delay: this should be a pain point. I hope it can be solved.

Can you provide more detail?
Are you talking about a physical mouse connected? Virtual mouse using the touchscreen?
General video delay?
What does the performance overlay say your network latency is?

BAByte:
@cgutman 
Thank you very much for your attention to this information:
1. Touch mode: You use RD Client and moonlight on ipad to make a product experience comparison: only use touch screen and keyboard to remotely control windows.
You will find that the experience of the two is very different. The experience of RD Client is very good, just like installing windows directly on the ipad.

2. Mouse delay: because my windows device is a notebook, and the graphics card only allows 30fps output stream. I changed the hdmi cheater to increase the refresh rate and solved it.

PR #629, opened 2024-05-21
Add passthrough touch support for windows native multi-touch

ZigZagT:
Adding support for sending touch events to Sunshine hosts directly.

This has been asked in multiple issues (incomplete list)

- https://github.com/moonlight-stream/moonlight-ios/issues/597
- https://github.com/moonlight-stream/moonlight-ios/issues/596
- https://github.com/moonlight-stream/moonlight-ios/issues/518

## Summary of changes

1. A new "passthrough touch events" option is added in settings when "Touchscreen" mode is on :

image

Once turned on, touch events will be routed to Sunshine hosts via the `LiSendTouchEvent` API, and appears as natively multi touch events on Windows. 

2. When passthrough mode is enabled, the pop-up keyboard would require *5 fingers* tap to fire up. This avoids conflicting with windows native gestures. (According to the [touch-gestures-for-windows](https://support.microsoft.com/en-us/windows/touch-gestures-for-windows-a9d28305-4818-a5df-4e2b-e5590f850741), windows has native touch gestures use up to 4 fingers, so 5 fingers tap provides a good guarantee). Windows also offers a built-in on-screen touch keyboard when multi-touch events are received, making the pop-up keyboard less crucial than before.

3. A performance tweak (turning off UIScrollView [delaysContentTouches](https://developer.apple.com/documentation/uikit/uiscrollview/1619398-delayscontenttouches?language=objc)). While I was on this I noticed the "pin & pan & scroll" feature in absolute touch mode brings in unwanted delay due to the default settings of ScrollView, which might have caused the following issues regarding delays:
- https://github.com/moonlight-stream/moonlight-ios/issues/496
- https://github.com/moonlight-stream/moonlight-ios/issues/489
The delay It is pretty noticeable in windows paint app.

### DIFF
--- a/Limelight/Input/PassthroughTouchHandler.m
+++ b/Limelight/Input/PassthroughTouchHandler.m
@@ -1,0 +1,64 @@
+//
+//  PassthroughTouchHandler.m
+//  Moonlight
+//
+//  Created by ZigZagT on 5/20/2024.
+//  Copyright © 2024 Moonlight Game Streaming Project. All rights reserved.
+//
+
+#import "PassthroughTouchHandler.h"
+
+#include <Limelight.h>
+
+@implementation PassthroughTouchHandler {
+    StreamView* view;
+}
+
+- (id)initWithView:(StreamView*)view {
+    self = [self init];
+    self->view = view;
+    return self;
+}
+
+- (int)sendTouchEvent:(UITouch*)touch withType:(uint8_t)eventType {
+    CGPoint location = [self->view adjustCoordinatesForVideoArea:[touch locationInView:self->view]];
+    CGSize videoSize = [self->view getVideoAreaSize];
+    return LiSendTouchEvent(
+        eventType,
+        (uint32_t) (uintptr_t) touch,
+        location.x / videoSize.width,
+        location.y / videoSize.height,
+        0,
+        0,
+        0,
+        LI_ROT_UNKNOWN
+    );
+}
+
+
+
+- (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
+    for (UITouch* touch in touches) {
+        [self sendTouchEvent:touch withType:LI_TOUCH_EVENT_DOWN];
+    }
+}
+
+- (void)touchesMoved:(NSSet *)touches withEvent:(UIEvent *)event {
+    for (UITouch* touch in touches) {
+        [self sendTouchEvent:touch withType:LI_TOUCH_EVENT_MOVE];
+    }
+}
+
+- (void)touchesEnded:(NSSet *)touches withEvent:(UIEvent *)event {
+    for (UITouch* touch in touches) {
+        [self sendTouchEvent:touch withType:LI_TOUCH_EVENT_UP];
+    }
+}
+
+- (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event {
+    for (UITouch* touch in touches) {
+        [self sendTouchEvent:touch withType:LI_TOUCH_EVENT_CANCEL];
+    }
+}
+
+@end
--- a/Limelight/Database/DataManager.h
+++ b/Limelight/Database/DataManager.h
@@ -28,6 +28,7 @@
                        enableHdr:(BOOL)enableHdr
                   btMouseSupport:(BOOL)btMouseSupport
                absoluteTouchMode:(BOOL)absoluteTouchMode
+            passthroughTouchMode:(BOOL)passthroughTouchMode
                     statsOverlay:(BOOL)statsOverlay;
 
 - (NSArray*) getHosts;
--- a/Limelight/Database/DataManager.m
+++ b/Limelight/Database/DataManager.m
@@ -67,6 +67,7 @@
                        enableHdr:(BOOL)enableHdr
                   btMouseSupport:(BOOL)btMouseSupport
                absoluteTouchMode:(BOOL)absoluteTouchMode
+            passthroughTouchMode:(BOOL)passthroughTouchMode
                     statsOverlay:(BOOL)statsOverlay {
     
     [_managedObjectContext performBlockAndWait:^{
@@ -86,6 +87,7 @@
         settingsToSave.enableHdr = enableHdr;
         settingsToSave.btMouseSupport = btMouseSupport;
         settingsToSave.absoluteTouchMode = absoluteTouchMode;
+        settingsToSave.passthroughTouchMode = passthroughTouchMode;
         settingsToSave.statsOverlay = statsOverlay;
         
         [self saveData];
--- a/Limelight/Database/TemporarySettings.h
+++ b/Limelight/Database/TemporarySettings.h
@@ -33,6 +33,7 @@
 @property (nonatomic) BOOL enableHdr;
 @property (nonatomic) BOOL btMouseSupport;
 @property (nonatomic) BOOL absoluteTouchMode;
+@property (nonatomic) BOOL passthroughTouchMode;
 @property (nonatomic) BOOL statsOverlay;
 
 - (id) initFromSettings:(Settings*)settings;
--- a/Limelight/Database/TemporarySettings.m
+++ b/Limelight/Database/TemporarySettings.m
@@ -84,6 +84,7 @@
     self.onscreenControls = settings.onscreenControls;
     self.btMouseSupport = settings.btMouseSupport;
     self.absoluteTouchMode = settings.absoluteTouchMode;
+    self.passthroughTouchMode = settings.passthroughTouchMode;
     self.statsOverlay = settings.statsOverlay;
 #endif
     self.uniqueId = settings.uniqueId;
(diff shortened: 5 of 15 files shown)


Limelight/Input/PassthroughTouchHandler.m
TrueZhuangJia, line 1: (uint32_t) (uint64_t) touch
Na, this could be buggy. I've been there. Easily causing pointer ID collision during intense touch operation, like a gaming scenario.  As 64bit mem addr is truncated into 32 bit, you have no idea if the lower 32bits repeats or not.
I use randomly generated pointer ID instead. CRC32 checksum also works better.

ZigZagT: can you clarify what "intense touch operation" means?

Most device support up to 10 points touches. I don't see a lot of opportunities for address collision. Mapping a 64 bit space to 32 bit naturally comes with the risk of colliding, and I don't see how CRC32 can make it better, not to mention the added overhead.

I do see a few things that might have give you this buggy impression:

- 64bit uint are not "truncated" into 32bit naively. Modulo arithmetic operations are used behind the doors per the C standard.
- It's undefined behavior if a pointer is casted to a narrower integer type.
- it's "implementation defined" behavior if signed integers are involved
- touch objects are only good for one touch sequence.  They should be cancel / ended promptly.

Debating aside, you're welcomed to test this branch with the "intense touch operation" settings.

ZigZagT: I guess using `uintptr_t` over `uint64_t` can make this more future proof. Tho it's a bit point less since we only have 64 bit devices these days.

TrueZhuangJia: "**Modulo arithmetic operations are used behind the doors**":
I have no idea how you found that in "C standard". Even it's true, Your approach is a convenient one but not a good idea for better coding since there's no "standard C", it all depends on your compiler and libraries.

Here's the fact: 
image
image
on my ipad mini6 iOS 17.2.
64bit integer gets simply truncated, naively.

TrueZhuangJia: **It's undefined behavior if a pointer is casted to a narrower integer type.**
_Yes, casting pointer to a narrower integer type is undefined, but it's also predictable on iOS._

it's "implementation defined" behavior if signed integers are involved
_Yes, by first converting pointer to a integer of equal length, make your approach a defined implementation, but it doesn't "justify" your approach._ 
Here's how:

image

TrueZhuangJia: **Debating aside, you're welcomed to test this branch with the "intense touch operation" settings.**

As I mentioned "I've been there", I'm not gonna test it since I've already kind of done that before. 
The original version of my fork was written by another Java guy who the code with me, he did a great job except for the pointer ID:
image
I demonstrated that above code results the same truncated integer just like your code.

I used his version of moonlight-iOS for native touch gameplay for hours long,  which is the "intense touch operation" I mentioned.
The buggy touch happens in an odd manner, it could be several times in a row or not triggered for minutes. You're not likely to reproduce it with some simple gesture touch like zoom in and out.

It was so frustrating that I decided to setup a macOS VM and review the code myself.
It took quite a long time before I noticed memory address is 64bit and the pointer ID requires 32 bit integer, and there's a chance of collision since upper 32 bits were discarded. After replacing pointer ID with CRC32 checksum of the 64bit address,  the bug was gone.

I'm not a veteran coding guy but all I said were based on fact oriented test and coding.

TrueZhuangJia: > I guess using `uintptr_t` over `uint64_t` can make this more future proof. Tho it's a bit point less since we only have 64 bit devices these days.

Indeed pointless. uintptr_t is the same stuff on 64 bit devices. If you read through my replies you will get to the point ---- my advice was based on long time multipoint touch gameplay testing.

ZigZagT: Like I listed in [my first reply](https://github.com/moonlight-stream/moonlight-ios/pull/629#discussion_r1611443544), there're many factors that can lead to the odd behavior you mentioned. 
I'd strongly recommend testing with "intensive touch" with this branch again to isolate the causes.

I've been casually using this build for a week now, and I have yet to discover any issues.

Without testing, I'm not convinced an at most 10 touch events can collide in 32 bit int space every 5-10 minutes and causing noticeable pain.

Adding additional crcsum is a significant maintenance burden that I'd prefer to steer clear of.

TrueZhuangJia: > Like I listed in [my first reply](https://github.com/moonlight-stream/moonlight-ios/pull/629#discussion_r1611443544), there're many factors that can lead to the odd behavior you mentioned. I'd strongly recommend testing with "intensive touch" with this branch again to isolate the causes.
> 
> I've been casually using this build for a week now, and I have yet to discover any issues.
> 
> Without testing, I'm not convinced an at most 10 touch events can collide in 32 bit int space every 5-10 minutes and causing noticeable pain.
> 
> Adding additional crcsum is a significant maintenance burden that I'd prefer to steer clear of.

Just a bit more fact: the gameplay scenario I mentioned involves only 2 fingers.  And buggy touch was gone after replacing truncated uint32_t with something else.  
I've mentioned all need to be said. However I understand truncated integer approach is good for you and we don't have to convince each other. I guess debate is over since it's becoming kind of pointless.
Thank you for your attention & discussion.

stefanilijev97:
This is really cool and long awaited feature 😀 when we can expect this to make it to the official releases?

testerl101:
Need help. This new option added isnt in the appstore release it seems, at least i dont have that option there. Because of this issue i can draw normally with stylus only if i enable touchpad instead of touchscreen. I am using an ipad pro 2025. Any way to get this option?

TrueZhuangJia:
> Need help. This new option added isnt in the appstore release it seems, at least i dont have that option there. Because of this issue i can draw normally with stylus only if i enable touchpad instead of touchscreen. I am using an ipad pro 2025. Any way to get this option?

I've review this PR, it's buggy on native touch pass through.
Mine has been verified to be reliable, also available on app store:
https://github.com/The-Fried-Fish/VoidLink-previously-moonlight-zwm

testerl101:
I will check user feeback on the stylus. Right now with moonlight it works fine if i disable touchscreen mode. The flicker is kinda buggy but doesnt cause any issues. If the styles is reported to be usually fine to use on your app, i will check it out.

usm916:
Great work on this feature — thanks for putting it together!  
While testing in OneNote’s drawing mode, I noticed a small issue: the first stroke works fine, but the second one often switches to text input or the hand (pan) tool.  
It seems that both stylus and mouse events are being sent during Apple Pencil hover, causing the cursor to flicker between the round stylus cursor and the arrow cursor.

### Tested Env
- **Client:** iPadOS / iPad Pro (M4) + Apple Pencil Pro  
- **App:** moonlight-ios (PR #629 branch)  
- **Host:** Windows 11 + Sunshine (multitouch enabled)  
- **Target App:** Microsoft OneNote (drawing mode)

### Steps to Reproduce / Expected vs Actual
1. Connect Moonlight iOS → open OneNote → enter drawing mode  
2. Draw the first stroke → start a second stroke  
- **Expected:** Continue drawing smoothly  
- **Actual:** Switches to text or pan mode; cursor flickers between stylus and arrow  
- **Frequency:** Consistent

### Cause (Hypothesis)
During Pencil hover, both stylus and mouse-like events are sent, which confuses Windows apps like OneNote and triggers mode changes.

### Proposed Fix (Minimal & Scoped)
When the Pencil is hovering or drawing, **prioritize stylus events** and **suppress mouse-like dispatch** during the hover period.  
Once the hover ends or input stops, allow mouse events again.

- **Detect:** Monitor Pencil hover and touch states (`UIHoverGestureRecognizer`, `UIPencilInteraction`, etc.)  
- **Gate:** Use `stylusHovering` / `stylusActive` flags to skip `LiSendMouse...` while preferring `LiSendTouch/Pen...`  
- **Scope:** Apply only when Passthrough Touch is enabled, to avoid affecting other modes

Reference (tested prototype):  
https://github.com/usm916/moonlight-ios/commit/5dbcbf018f7c7a1d24fee1208ed651774f6c5218

### Validation Results
- Cursor flicker during hover is gone or greatly reduced  
- Continuous drawing works normally  
- Mouse input unaffected when Pencil is away  
- No change when Passthrough Touch is off

### Follow-ups
Would you be open to adding this as a small follow-up commit to PR #629? 

I'm still a beginner with OSS contributions, so please let me know if I can improve how I share feedback.

ZigZagT:
I'm going to close this since this repo is seemingly abandoned by the maintainers. I'm maintaining my fork for my own use so feel free to pull the source and build it yourself.

stefanilijev97:
> I'm going to close this since this repo is seemingly abandoned by the maintainers. I'm maintaining my fork for my own use so feel free to pull the source and build it yourself.

It's better to disregard this repo and use VoidLink (available on AppStore), which is basically Moonlight on steroids and has really nice features, such as: touch improvements, apple pen support, better streaming options etc.

testerl101:
Yeah and it charges you like 3x base price on top of the base price just to use drawing features, no thank you.

PR #646, opened 2024-09-07
Fixing a crash related to the OpenSSL library

sergeishere:
Making cert with OpenSSL from the latest commit causes a crash because it fails to load the legacy provider. To avoid the crash and allow legacy algorithms, the library needed to be rebuilt.

I rebuilt the libraries with the necessary parameters and added them to the repository. During the build, I disabled unused algorithms and technologies to reduce the size. The libraries are linked in `Build Settings` through `Other Linker Flags`.

Parameters using during build:

```
enable-weak-ssl-ciphers enable-ssl2 no-asm no-shared no-module no-sock no-hw no-afalgeng no-apps no-async no-capieng no-comp no-deprecated no-dgram no-dso no-gost no-http no-nextprotoneg no-ocsp no-padlockeng no-sm2-precomp no-srp no-srtp no-sse2 no-ssl-trace no-quic no-ui-console no-uplink no-tls no-dtls no-tests no-aria no-bf no-blake2 no-camellia no-cast no-chacha no-cmac no-dh no-dsa no-ecdh no-ecdsa no-idea no-md4 no-mdc2 no-ocb no-poly1305 no-rc4 no-rmd160 no-scrypt no-seed no-siphash no-siv no-sm2 no-sm3 no-sm4 no-whirlpool no-ec no-ec2m
```

All schemes have been tested except for the Apple TV device (I don't have one).

The external dependency from the previous commit was accordingly removed, as it does not contain correct binaries.


Priultimus:
I run into this error when I attempt to compile it on Xcode 16 -- not sure how if this is my own configuration issue or not though
screenshot3

sergeishere:
@Priultimus It seems that there are build artifacts left from the previous version — the screenshot mentions OpenSSL.framework, which should no longer be there. Can you try to clean the build folder for the main project and `moonlight-common` (or clean up the DerivedData folder) and then rebuild the project.

sergeishere:
I'm also having issues with the project not building in Xcode 16 RC. Neither the main branch nor my openssl-fix branch.

Priultimus:
On your branch Xcode lets me edit crypto.h, presumably because OpenSSL is precompiled and in this case if you just actually do what the IDE says and add "#include <sys/_types/_time_t.h>" to crypto.h it appears like everything just works lol

anastasiuspernat:
I was able to compile with XCode 16 upgrading OpenSSL to v3.3.2000 but now it crashes during run time.

PR #664, opened 2024-12-23
Support for Device gyro instead of controller gyro via CoreMotion

Danos0100:
This PR lets the user choose if they want to use Controller Motion or Device Motion (so you can play with motion without controller/with a non-motion-enabled controller)

As this requires CMAcceleration and CMRotation instead of GCAcceleration and GCRotation, this makes the setMotionEventState function a little bit longer.

Also, this adds a UI Switch to change the new motionMode variable for player 0/1:

- _Auto_ uses device motion as a fallback if controller motion is not available
- _Device Motion_ always uses device motion sensors (this possibillity is needed, as some controller grips send useless dummy motion data to the iPhone. In this case, the user can manually select to use the device's motion data instead.)
- _Controller Motion_ always uses the controller's motion data. If the controller doesn't offer motion data, the device is **not** used as a fallback.

### DIFF
--- a/Limelight/Database/DataManager.h
+++ b/Limelight/Database/DataManager.h
@@ -19,6 +19,7 @@
                            width:(NSInteger)width
                 audioConfig:(NSInteger)audioConfig
                 onscreenControls:(NSInteger)onscreenControls
+                      motionMode:(NSInteger)motionMode
                    optimizeGames:(BOOL)optimizeGames
                  multiController:(BOOL)multiController
                  swapABXYButtons:(BOOL)swapABXYButtons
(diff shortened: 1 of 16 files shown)


Danos0100:
(For those who want to compile it for themselves, I suggest to use this fork instead. It also incorporates the openssl-fix by @sergeishere: https://github.com/Danos0100/moonlight-ios/tree/devicemotion-support)

seastwood:
Thank you for this. Went through the trouble of building and installing this branch onto my iphone 15 plus. Works great!

jfoxwoosh:
any chance of this getting merged?

PR #671, opened 2025-05-07
[opt] Use UILabel for stats overlay

thieman:
As part of a general performance investigation, I found out that rendering the stats overlay was taking up to 20ms on my ATV 4K 1st Gen. This hangs the main thread long enough that the next displayLinkCallback is missed and results in skipped frames. A snapshot from a profile below shows a spike where the main thread was saturated due to activity related to `updateOverlayText`.

Pasted Graphic

Our goal here is to reduce the time spent drawing the stats overlay. To do this, we switch the overlay from a UITextView to a UILabel. I am a complete novice at iOS, I just tried this because the UILabel seemed like a simpler solution for rendering a simple stats overlay. However, this did seem to improve the rendering times by quite a bit. On the same device, I'm seeing `updateOverlayText` typically take only around 3ms now, down from 20ms. This has removed the visible frame skips that previously happened with the stats overlay is displayed.

As part of this, also took another stab at styling the overlay since the default UILabel was pretty ugly. Added a bit of margin at the top, increased opacity slightly, and increased font weight. I think the new label is much more legible but you can still see through to the content if you need to.

Screenshot of the new label

image

### DIFF
--- a/Limelight/Utility/PaddedLabel.h
+++ b/Limelight/Utility/PaddedLabel.h
@@ -1,0 +1,15 @@
+//
+//  PaddedLabel.h
+//  Moonlight
+//
+//  Created by Travis Thieman on 5/6/25.
+//  Copyright © 2025 Moonlight Game Streaming Project. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface PaddedLabel : UILabel
+
+@property (nonatomic, assign) UIEdgeInsets textInsets;
+
+@end
--- a/Limelight/ViewControllers/StreamFrameViewController.m
+++ b/Limelight/ViewControllers/StreamFrameViewController.m
@@ -12,6 +12,7 @@
 #import "StreamManager.h"
 #import "ControllerSupport.h"
 #import "DataManager.h"
+#import "PaddedLabel.h"
 
 #include <sys/socket.h>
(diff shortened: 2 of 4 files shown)


Limelight/Utility/PaddedLabel.h
```diff
@@ -0,0 +1,15 @@
+//
```
thieman, line 1: Disclosure: I am a complete iOS novice, some of this PR was generated by LLM tools but I've read through everything submitted here.

Limelight/ViewControllers/StreamFrameViewController.m
```diff
@@ -297,7 +302,7 @@ - (void)updateOverlayText:(NSString*)text {
                                            _overlayView.frame.size.height)];
         [_overlayView setText:text];
         [_overlayView sizeToFit];
-        [_overlayView setCenter:CGPointMake(self.view.frame.size.width / 2, _overlayView.frame.size.height / 2)];
+        [_overlayView setCenter:CGPointMake(self.view.frame.size.width / 2, (12 + (_overlayView.frame.size.height / 2)))];
```
thieman, line 301: I'm guessing having a random `12` in here is not the best idea, is there a better way I could do this?

Issue #675, opened 2025-07-22
Mouse speed way too slow on iPad


MadhanMurali:
**Describe the bug**
The mouse cursor in moonlight moves very slow when I use trackpad of the keyboard case I mentioned, even when ipad's trackpad and pc's mouse sensitivity are maxed out. I tried using xrdp and gnome remote desktop, and mouse cursor speed is very good and works as expected.

**Steps to reproduce**
Run sunshine and connect to PC using iPad and a keyboard with a trackpad (not sure this happens with magic keyboard). And then try moving the cursor.

**Screenshots**
Not Applicable

**Affected games**
Not Applicable

**Other Moonlight clients**
Not Applicable

**Moonlight settings (please complete the following information)**
<img width="1411" height="1908" alt="Image" src="https://github.com/user-attachments/assets/f7b1e695-d9e3-40dd-aa5e-a734a82ecc4c" />

**Gamepad-related issues (please complete if problem is gamepad-related)**
Not Applicable

**Device details (please complete the following information)**
Hardware:
- iPad Air M2 11 inch
- Typecase keyboard - https://typecase.in/product/1395/

Software:
- IPadOS 18.5
- Current version of moonlight and sunshine

**Server PC details (please complete the following information)**
 - OS: Debian 12
 - GeForce Experience version: N/A
 - Nvidia GPU driver: 535.247.01
 - Antivirus and firewall software: N/A

**Additional context**
I have clipped the screen to fit ipad's screen using nvidia-settings using this command: nvidia-settings --assign CurrentMetaMode="DP-0: 2560x1440 { ViewPortIn=2072x1440, ViewPortOut=2072x1440+244+0 }"
But even if i skip doing this, the issue persists.

Edit: Relevant issue?: https://github.com/moonlight-stream/moonlight-ios/issues/491

Elaws:
Hi, any news please ?

PR #689, opened 2025-12-04
Increase mouse speed

hexhexD:
I'm using moonlight iPad daily, and the magic keyboard trackpad is slow.

https://github.com/moonlight-stream/moonlight-ios/issues/675


crescentduchess:
Having the same issue using Sunshine. Trackpad and mouse both far too slow with iOS client.

enekodr:
Anyone can confirm if this happens in AppleTV as well? Looks like this change may affect other clients

